Issue #72, opened 2021-10-26
[BUG] toplogy does not load when loading console for the first time


jo19in1:
**Describe the bug**
When a signed-in user loads the console on the browser, the topology view does not show up

**To Reproduce**
1. Go to <url> and sign in. Now close the tab.
2. Open a fresh browser window and load <url/#/topology>
3. You will not see the toplogy graphs.
4. Below is a screenshot of the same 
![image](https://user-images.githubusercontent.com/34362153/138811417-86161e45-dfb3-4ddf-9d26-59bcf8521de8.png)

5. Refresh the page and you will see the toplogy
![image](https://user-images.githubusercontent.com/34362153/138811524-cbc20cda-0ade-416d-9d5a-a2308617acb1.png)

**Expected behavior**

**Screenshots**

**Components/Services affected**

- [x] UI/Frontend
- [ ] API/Backend
- [ ] Agent
- [ ] Deployment/YAMLs
- [ ] CI/CD Integration
- [ ] Other (specify)

**Additional context**

PR #90, opened 2021-11-02
fix refresh token flow on initial load

manV:
ref: https://github.com/deepfence/ThreatMapper/issues/72

### DIFF
--- a/deepfence_ui/app/scripts/helpers/auth-helper.js
+++ b/deepfence_ui/app/scripts/helpers/auth-helper.js
@@ -34,7 +34,7 @@ export function isUserSessionActive() {
   const authToken = localStorage.getItem('authToken');
   if (authToken) {
     setInterval(() => {
-      startTimerForRefreshToken();
+      refreshAuthTokenIfRequired();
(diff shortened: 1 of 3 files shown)


deepfence_ui/app/scripts/helpers/auth-helper.js
```diff
@@ -51,17 +51,39 @@ export function isUserSessionActive() {
...
+    setInterval(() => {
+      refreshAuthTokenIfRequired();
+    }, 60 * 5 * 1000);
+    await refreshAuthTokenIfRequired();
+  }
+  if (authToken) {
+    isSessionActive = true;
+  } else {
+    isSessionActive = false;
+  }
```
ibreakthecloud, line 67: can this be minified?
```suggestion
  if (authToken) {
    setInterval(() => {
      refreshAuthTokenIfRequired();
    }, 60 * 5 * 1000);
    await refreshAuthTokenIfRequired();
    isSessionActive = true;
  }
```

ibreakthecloud (approved):

PR #208, opened 2022-01-10
Vulnerability page download fix

saiprasanth1303:
- Fixed the vulnerability page XLSX download action error that was caused due to the reports merge

@deepfence/engineering

### DIFF
--- a/deepfence_ui/app/scripts/components/vulnerability-view/cve-image-report-container.js
+++ b/deepfence_ui/app/scripts/components/vulnerability-view/cve-image-report-container.js
@@ -44,14 +44,9 @@ class CVEImageReportContainer extends React.PureComponent {
     dispatch(toaster('Generating Report ...'));
     generateReport.then(res => {
       if (res.success) {
-        console.log('Generating report is done, Fetching file');
         const fetchReportStatus = dispatch(reportDownloadStatusAction());
         fetchReportStatus.then(async res => {
           if (res.data[0]) {
-            console.log(
-              'Status fetch sucessful, Report:',
-              res.data[0].report_id
-            );
             await new Promise(resolve => setTimeout(resolve, 1000));
             const downloadReport = dispatch(
               downloadReportAction({
--- a/deepfence_ui/app/scripts/components/vulnerability-view/cve-image-report-container.js
+++ b/deepfence_ui/app/scripts/components/vulnerability-view/cve-image-report-container.js
@@ -1,10 +1,12 @@
+/* eslint-disable */
 import React from 'react';
-import {connect} from 'react-redux';
+import { connect } from 'react-redux';
+import { nodeFilterValueSelector } from '../../selectors/node-filters';
 import {
-  nodeFilterValueSelector,
-} from '../../selectors/node-filters';
-import {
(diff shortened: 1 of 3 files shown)


deepfence_ui/app/scripts/components/vulnerability-view/cve-image-report-container.js
```diff
@@ -14,46 +16,65 @@ class CVEImageReportContainer extends React.PureComponent {
     this.handleDownload = this.handleDownload.bind(this);
   }
 
-  handleDownload(params) {
-    const {
-      xlsxReportDownloadAction: action,
-    } = this.props;
+  async handleDownload(params) {
+    const { dispatch } = this.props;
```
manV, line 20: why take dispatch in props? when we can use `useDispatch`?

deepfence_ui/app/scripts/components/vulnerability-view/cve-image-report-container.js
```diff
@@ -14,46 +16,65 @@ class CVEImageReportContainer extends React.PureComponent {
...
 
-    return action(apiParams);
+    const generateReport = dispatch(reportGenerateAction(apiParams));
+    dispatch(toaster('Generating Report ...'));
+    generateReport.then(res => {
+      if (res.success) {
+        const fetchReportStatus = dispatch(reportDownloadStatusAction());
+        fetchReportStatus.then(async res => {
+          if (res.data[0]) {
+            await new Promise(resolve => setTimeout(resolve, 1000));
```
manV, line 50: why artificial delay?

saiprasanth1303: I added this because the report is taking at least a second to fully generate before we can download it. If the delay isn't there, the report is not getting downloaded because it is not fully generated yet.

deepfence_ui/app/scripts/components/vulnerability-view/cve-image-report-container.js
saiprasanth1303: These files are still class based. They are part of the list which needs to be refactored. Once this fix is pushed, I'll refactor them to functional components.

manV (approved):

deepfence_ui/app/scripts/components/vulnerability-view/cve-image-report-container.js
ramanan-ravi: @saiprasanth1303 I think hardcoding 2 second wait won't work in all cases. Report generation time is dependent on the amount of data we process / no of pages. So it can take > 2 seconds if the image / host has 1000's of vulnerabilities.
Can you add 3 retries and then throw error?

saiprasanth1303: @ramanan-ravi Shall I add the 3 retries after the 2-second wait?

ramanan-ravi: Yes. Also 2 second wait between retries.

shyam-dev:
Can we make this event based instead of doing a timed wait ?

PR #235, opened 2022-01-27
Fix issue with toplogy graph height cropping

manV:
- replaces react-tippy which was messing with body height with official tippyjs react library.

@deepfence/engineering


pandyamarut (approved):
/LGTM

PR #357, opened 2022-03-22
donut modal popup severity sorting fix

pandyamarut:
Signed-off-by: Marut Pandya <pandyamarut@gmail.com>

@deepfence/engineering

### DIFF
--- a/deepfence_ui/app/scripts/components/topology-view/donut-details-modal/donut-details-modal.js
+++ b/deepfence_ui/app/scripts/components/topology-view/donut-details-modal/donut-details-modal.js
@@ -1,3 +1,4 @@
+/* eslint-disable class-methods-use-this */
 // React imports
 import React, { useCallback, useMemo, useState } from 'react';
 import { connect } from 'react-redux';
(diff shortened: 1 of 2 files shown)


deepfence_ui/app/scripts/components/topology-view/donut-details-modal/donut-details-modal.js
```diff
@@ -164,6 +169,19 @@ class DonutDetailsModal extends React.Component {
...
+    if(data[0]?.desc === true) {
+      this.setState({ sortSeverity: 'desc'})
+    } else if (data[0]?.desc === false) {
+      this.setState({sortSeverity: 'asc'})
+    } else {
+      this.setState({ sortSeverity: 'desc'})
+    }
+    setTimeout(() => {
+      this.fetchAlertsDetails();
+    }, 0);
```
manV, line 182: we need to check which column is getting sorted here. and apply sorting only if severity column is being sorted. the logic to set state and trigger api call can be made better with something like

```suggestion
    this.setState({
      sortSeverity: !data[0]?.desc ? 'asc' : 'desc' 
    }, () => {
      this.fetchAlertsDetails();
    });
```

manV: We should enable sorting for only severity column, and disable it on all other columns. having `enableSorting` true on table means all the columns will be sortable.

pandyamarut: @manV  right. 
I will do the changes. Thanks.

manV (approved):

PR #411, opened 2022-04-21
Add support to upload db files in airgapped enviornment

pandyamarut:
Signed-off-by: Marut Pandya <pandyamarut@gmail.com>

Issue: https://github.com/deepfence/ThreatMapper/issues/366

Issue #366, opened 2022-03-30
Vulnerability database: Support offline and air-gapped environments


ramanan-ravi:
- We need to support vulnerability scans in offline and air-gapped environments
- User should be able to download the feed from our server and upload it in the management console via UI
- This will eliminate the need for internet access in those environments

PR #412, opened 2022-04-21
(feat). API to upload vulnerability db in airgapped env

ibreakthecloud:
API: `df-api/upload-vulnerability-db`

Usage:
```
curl --location --request POST 'https://my-console.df.io/df-api/upload-vulnerability-db' \
--header 'Content-Type: application/json' \
--header 'deepfence-key: xxxxxxxxxx-xxxx-xxxxxxxxxx-xxx' \
--form 'file=@"/Users/karn/Downloads/threatintel-vuln-1648219229250.tar.gz"'
```
Changes:
- add api in fetcher to receive `.tar.gz` size less than 150MB
- add api in vulnerability-mapper that updates vulnerability db

Related PR in UI: #411 
Issue: https://github.com/deepfence/ThreatMapper/issues/366

### DIFF
--- a/deepfence_console/fetcher/fetcher-server.go
+++ b/deepfence_console/fetcher/fetcher-server.go
@@ -502,7 +502,7 @@ func handleVulnerabilityFeedTarUpload(respWrite http.ResponseWriter, req *http.R
 		vulnerabilityDbUpdater.updateVulnerabilityDbListing()
 
 		// call mapper api to update grype db
-		updateVulnerabilityMapperDB(respWrite)
+		updateVulnerabilityMapperDB()
 	}()
 	_, err = fmt.Fprintf(respWrite, "vulnerability db updated")
 	if err != nil {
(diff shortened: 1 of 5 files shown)


deepfence_console/fetcher/fetcher-server.go
```diff
@@ -445,6 +445,82 @@ func checkOwaspDependencyDataDownloading() bool {
...
+		}
+		vulnerabilityDbUpdater.updateVulnerabilityDbListing()
+	}()
+	// }
+	_, err = fmt.Fprintf(respWrite, "POST complete")
+	if err != nil {
+		fmt.Println(handler.Filename, err)
+	}
+
+	// call mapper api to update grype db
```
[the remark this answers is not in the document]
ibreakthecloud: makes sense. updated

deepfence_console/fetcher/fetcher-server.go
```diff
@@ -445,6 +445,90 @@ func checkOwaspDependencyDataDownloading() bool {
...
+	}()
+	_, err = fmt.Fprintf(respWrite, "vulnerability db updated")
+	if err != nil {
+		fmt.Println(handler.Filename, err)
+	}
+
+	respWrite.WriteHeader(http.StatusOK)
+}
+
+func updateVulnerabilityMapperDB(respWrite http.ResponseWriter) {
```
ramanan-ravi, line 515: `respWrite` should not be passed here since it is running in a goroutine.
Since we have already sent 200 response at line 512, we should not be writing anything to `respWrite` anymore I think. If there are any errors we can log to stderr.

ibreakthecloud: pushed pls. check

PR #449, opened 2022-05-26
[Iss1510] Data under Sbom is not shown right

mukuldeepfence:
Fixes # Redundant data under runtimeBOM has been fixed

- when we select filters using lucene string few BOM were getting appeared other than runtimeBOM 

@deepfence/engineering


mukuldeepfence:
Please merge iss1503 and then iss1510 branch to TM

PR #478, opened 2022-06-30
Fix connections report

noboruma:
Fixes missing connections report.

- Avoid race condition on non-eBPF route by removing global variables
- Make sure kprobes are correctly cleaned up if loading them failed once 

@deepfence/engineering

### DIFF
--- a/deepfence_agent/tools/apache/scope/vendor/github.com/weaveworks/tcptracer-bpf/pkg/tracer/tracer.go
+++ b/deepfence_agent/tools/apache/scope/vendor/github.com/weaveworks/tcptracer-bpf/pkg/tracer/tracer.go
@@ -1,3 +1,4 @@
+//go:build linux
 // +build linux
 
 package tracer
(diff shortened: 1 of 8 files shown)


ramanan-ravi (approved):

deepfence_agent/tools/apache/scope/vendor/github.com/weaveworks/tcptracer-bpf/pkg/tracer/tracer.go
```diff
@@ -31,26 +32,46 @@ func TracerAsset() ([]byte, error) {
...
-	sectionParams := make(map[string]bpflib.SectionParams)
-	sectionParams["maps/tcp_event_ipv4"] = bpflib.SectionParams{PerfRingBufferPageCount: 256}
-	err = m.Load(sectionParams)
-	if err != nil {
-		return nil, err
+		err = m.EnableKprobes(maxActive)
+		if err == nil {
+			break
+		} else {
+			if retry == 0 {
```
shyam-dev, line 60: The retry value is currently set to 1. Which means that this loop will be executed only once ! I think we should set the retry value to a larger number with some wait time between the retries. Or we should remove this piece of code.

PR #479, opened 2022-07-05
Graph style improvments

manV:
- Graph on pop up has matching background
- Improved topology graph edge styles

@deepfence/engineering


manV:
Similar changes merged with Attack Graph.

PR #512, opened 2022-08-03
Fixed user-sent notifications

jilleJr:
- Fixed bug where user-sent notifications resulted in Python exception

I've no idea what values you all want in these params, so I'm just guessing. I made these changes inside the running container and now I can send notifications through the `/#/vulnerability/details/{docker-image-name-and-tag}` page by clicking the checkbox on a CVE and then the bell ("Notify")

Had to dig through the code a bit to understand, because the only thing that was logged out of the `deepfence-api` was:

```log
[2022-08-03 14:08:32 +0000] [929] [ERROR] Error sending notification: 'notification_id'
ERROR:config.app:Error sending notification: 'notification_id'
```

@deepfence/engineering


ramanan-ravi:
Hello @jilleJr , thanks for the contribution.
We fixed the issue [here](https://github.com/deepfence/ThreatMapper/pull/555) since some more changes were required.

PR #536, opened 2022-08-22
Discrepancy in the results while we select container image in report download

mukuldeepfence:
Fixes # Discrepancy in the results while we select container image node_type and the container image name has been fixed

- when we select specific image to get results we should be able to see only results related to that image only

@deepfence/engineering

### DIFF
--- a/deepfence_backend/utils/esconn.py
+++ b/deepfence_backend/utils/esconn.py
@@ -1356,7 +1356,6 @@ class ESConn:
                            lucene_query_string=None, add_masked_filter=True, get_only_query=False):
         should_objects = []
         range_query = None
-        print(index_name, filters, aggs, number, time_unit)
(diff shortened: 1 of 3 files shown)


ramanan-ravi (approved):

deepfence_backend/utils/esconn.py
```diff
@@ -1356,7 +1356,9 @@ def aggregation_helper(index_name, filters, aggs, number=None, time_unit=None,
                            lucene_query_string=None, add_masked_filter=True, get_only_query=False):
         should_objects = []
         range_query = None
+        print(index_name, filters, aggs, number, time_unit)
```
ramanan-ravi, line 1359: can you remove debugging statements

mukuldeepfence: yes i have removed @ramanan-ravi

PR #544, opened 2022-08-31
UI: Show checkbox for only relevant node types on topology table

manV:
- Show checkbox for selection for only relevant nodes. Hide it for nodes like cloud, regions etc where there are no actions available.

@deepfence/engineering


milan-deepfence:
looks good to me

PR #562, opened 2022-09-13
Cloudtrail alerts UI

saiprasanth1303:
Fixes # .

- Added cloudtrail integration options to the Integrations 

@deepfence/engineering

### DIFF
--- a/deepfence_ui/app/scripts/components/integration-view/advance-filter-modal.js
+++ b/deepfence_ui/app/scripts/components/integration-view/advance-filter-modal.js
@@ -63,9 +63,12 @@ export const AdvancedFilterModalContent = ({
 
   const filterValues = filters || initialFilters;
 
+  // eslint-disable-next-line array-callback-return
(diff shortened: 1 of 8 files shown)


deepfence_ui/app/scripts/components/integration-view/advance-filter-modal.js
```diff
@@ -63,9 +63,12 @@ export const AdvancedFilterModalContent = ({
 
   const filterValues = filters || initialFilters;
 
+  // eslint-disable-next-line array-callback-return
+  const nodeFiltersOptions = nodeFilters && nodeFilters.filter(item => { if (item.label !== 'CloudTrail') return item;});
```
manV, line 67: I don't get this, filter callback is supposed to return boolean.

saiprasanth1303: I added this because we don't want to show the cloudtrail filter on the advanced filters popup since the cloudtrail filter comes through the enumerate API. There is a check now that hides the advanced filters button when the user selects cloudtrail under resource. But it would still show the filter in the advanced filters pop-up when the user selects any other resource. This line removes the cloudtrail filter just for the popup

PR #593, opened 2022-09-27
Fixed empty option value in the table column options

saiprasanth1303:
Fixes # .

- Fixed empty option value in the table column options

@deepfence/engineering

### DIFF
--- a/deepfence_ui/app/scripts/components/common/multi-select/app-trigger.js
+++ b/deepfence_ui/app/scripts/components/common/multi-select/app-trigger.js
@@ -67,7 +67,6 @@ const Dropdown = (props) => {
 
 const Option = props => (
   <div>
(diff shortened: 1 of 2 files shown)


manV (changes requested):
I think instead of changing the dropdown component itself, we should fix the table in question. Currently we are setting empty header for the column we want to hide instead of removing it from the columns array altogether, we should change this and remove the column from columns array if we don't want to show this.

saiprasanth1303:
@manV We can't remove that column because we still need to show it for cloud scan results. There is no data for it in host or cluster scans so we just hide it by not showing the column name.

manV:
@saiprasanth1303 why can't we remove column conditionally?

manV:
Stale, Closing.

PR #720, opened 2022-12-02
Optimise new UI dockerfile

manV:
- optimise new UI dockerfile


deepfence_frontend/Dockerfile
```diff
@@ -1,11 +1,17 @@
-ARG DF_IMG_TAG=latest
```
manV, line 1: @ramanan-ravi I removed these lines, not sure what they were for

PR #723, opened 2022-12-02
added to slide modal from left

milan-deepfence:
- Open modal from left option added

### DIFF
--- a/deepfence_frontend/packages/ui-components/src/components/avatar/Avatar.tsx
+++ b/deepfence_frontend/packages/ui-components/src/components/avatar/Avatar.tsx
@@ -7,7 +7,6 @@ import { twMerge } from 'tailwind-merge';
 import { Typography } from '@/components/typography/Typography';
 
 type AvatarType = {
-  asChild?: boolean;
   alt?: string;
   src?: string;
   className?: string;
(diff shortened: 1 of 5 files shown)


deepfence_frontend/packages/ui-components/src/components/avatar/Avatar.tsx
```diff
@@ -32,28 +33,37 @@ const Child = ({ children }: { children: AvatarType['children'] }) => {
...
-      )}
-    >
-      {!asChild ? <img src={src} alt={alt} className="p-2" /> : <Child>{children}</Child>}
-    </button>
-  );
-};
+export const Avatar = forwardRef<HTMLButtonElement, AvatarType>(
+  (
+    {
+      asChild = false,
```
manV, line 39: why are we taking boolean `asChild` along with `children`. If children is `undefined` we can render icon, otherwise render `children`.

manV (approved):

deepfence_frontend/packages/ui-components/src/components/avatar/Avatar.tsx
milan-deepfence: Remove asChild props, if require more things we will update later.

milan-deepfence: Remove asChild props, if require more things we will update later.

PR #831, opened 2023-01-19
Cache kubernetes report and generate once every 5 minutes

### LINKED ISSUE #777: Kubernetes reporter issue in large clusters
Generating kubernetes report in large k8s cluster takes longer than report interval. This results in inconsistent data.

```
k8s reporter took 19s (longer than 5s)
```

ramanan-ravi:
Fixes #777

### DIFF
--- a/deepfence_agent/tools/apache/scope/probe/kubernetes/reporter.go
+++ b/deepfence_agent/tools/apache/scope/probe/kubernetes/reporter.go
@@ -3,6 +3,10 @@ package kubernetes
 import (
 	"os"
 	"strings"
+	"sync"
+	"time"
+
+	"github.com/sirupsen/logrus"
 
 	"k8s.io/apimachinery/pkg/labels"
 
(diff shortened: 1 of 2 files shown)


deepfence_agent/tools/apache/scope/probe/kubernetes/reporter.go
```diff
@@ -346,8 +358,41 @@ func (r *Tagger) Tag(rpt report.Report) (report.Report, error) {
 	return rpt, nil
 }
 
-// Report generates a Report containing Container and ContainerImage topologies
+func (r *Reporter) updateReportCache() {
+	k8sReport, err := r.generateReport()
+	if err != nil {
+		logrus.Error(err.Error())
```
shyam-dev, line 364: If we get an error, shouldn't we return ? Or do something else ?

deepfence_agent/tools/apache/scope/probe/kubernetes/reporter.go
```diff
@@ -346,8 +358,41 @@ func (r *Tagger) Tag(rpt report.Report) (report.Report, error) {
...
+	} else {
+		r.kubernetesReportCache.Lock()
+		r.kubernetesReportCache.k8sReport = k8sReport
+		r.kubernetesReportCache.Unlock()
+	}
+	ticker := time.NewTicker(5 * time.Minute)
+	for {
+		select {
+		case <-ticker.C:
+			k8sReport, err = r.generateReport()
```
shyam-dev, line 374: Will this function complete before 5* time.Minute all the time ? Are there cases where it will be slow ? Also, what happens if this function always returns error ? Is there a way to notify, bail out and retry after sometime ?

deepfence_agent/tools/apache/scope/probe/kubernetes/reporter.go
ramanan-ravi: This is first report collection, on start up. If it is a temporary error, it will be resolved in the next generated report (after 5 minutes)

deepfence_agent/tools/apache/scope/probe/kubernetes/reporter.go
ramanan-ravi: If the report generation is slower than 5 minutes, it won't be an error. We would have just kicked off next report generation. We are safe in terms of updating the cache because of lock, in case those two simultaneous reports are finished at the same time.

PR #833, opened 2023-01-20
Display image name in details page, sbom header instead of image id

milan-deepfence:
- To display image name instead of image id for vulnerability

@deepfence/engineering

### DIFF
--- a/deepfence_ui/app/scripts/components/common/detail-modal/header.jsx
+++ b/deepfence_ui/app/scripts/components/common/detail-modal/header.jsx
@@ -10,7 +10,7 @@ const HeaderKVPair = ({ dataKey, value, valueAsText }) => {
   return (
     <div className={styles.alertSummaryItem}>
       <div className={styles.alertSummaryItemTitle}>{dataKey}</div>
-      <div className={styles.alertSummaryItemContent} title={valueAsText}>{value}</div>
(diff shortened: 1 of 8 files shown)


deepfence_ui/app/scripts/components/common/detail-modal/header.jsx
```diff
@@ -10,7 +10,7 @@ const HeaderKVPair = ({ dataKey, value, valueAsText }) => {
   return (
     <div className={styles.alertSummaryItem}>
       <div className={styles.alertSummaryItemTitle}>{dataKey}</div>
-      <div className={styles.alertSummaryItemContent} title={valueAsText}>{value}</div>
+      <div className={styles.alertSummaryItemContent} title={valueAsText}>{valueAsText}</div>
```
manV, line 13: is this needed?

manV (approved):

deepfence_ui/app/scripts/components/common/detail-modal/header.jsx
milan-deepfence: Yeah, valueAsText is needed. Value is being displayed in ui which carry node_id(image_id) and is used to trigger api.
valueAsText is used to display to only ui not use for any logical things.

PR #875, opened 2023-02-08
[V2][UI] Add registries support for onboard connectors list

manV:
- Add registries support for onboard connectors list

### DIFF
--- a/deepfence_frontend/apps/dashboard/src/utils/registry.ts
+++ b/deepfence_frontend/apps/dashboard/src/utils/registry.ts
@@ -1,0 +1,17 @@
+import { isNil } from 'lodash-es';
+
+import { ModelRegistryListResp } from '@/api/generated';
+
+export function getRegistryDisplayId(registry: ModelRegistryListResp): string {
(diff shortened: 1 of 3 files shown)


deepfence_frontend/apps/dashboard/src/utils/registry.ts
```diff
@@ -0,0 +1,17 @@
+import { isNil } from 'lodash-es';
+
+import { ModelRegistryListResp } from '@/api/generated';
+
+export function getRegistryDisplayId(registry: ModelRegistryListResp): string {
+  if (
+    registry.registry_type === 'docker_hub' &&
```
milan-deepfence, line 7: Any reason why just docker_hub?

milan-deepfence (approved):

deepfence_frontend/apps/dashboard/src/utils/registry.ts
manV: that's what we have currently, we will add more as api support is added.

PR #902, opened 2023-02-15
V2 resource ingest

deviprasad303:
Fixes # .

-
-
-

@deepfence/engineering


deepfence_server/ingesters/cloud_resource_ingester.go
```diff
@@ -100,6 +110,14 @@ func (tc *CloudComplianceIngester) LinkNodesWithCloudResources(ctx context.Conte
 		return err
 	}
 
+	if _, err = tx.Run("match (n:Node) WITH apoc.convert.fromJsonMap(n.cloud_metadata) as map, n WHERE map.label = 'GCP' WITH map.hostname as hostname, n match (m:CloudResource) where m.resource_type = 'gcp_compute_instance' and m.hostname = hostname MERGE (n) -[:IS]-> (m)", map[string]interface{}{}); err != nil {
```
noboruma, line 113: nit: Instead of using `"query"` we can use ``
That way we can format the query a bit better and make it readable, like:
```
`
MATCH (n)
RETURN n
`
```

PR #950, opened 2023-03-21
added subheader in posture scan result page

milan-deepfence:
- Added sub header fixed component layout

@deepfence/engineering


manV:
We will pursue other ways of doing this

Issue #725, opened 2022-12-05
[Graph integration] Add new deepfence server


noboruma:
As part of the ongoing effort to integrate with neo4j graph (https://github.com/deepfence/ThreatMapper/issues/692) we are unifying & revamping our main web server.

- [x] Add unified http server
- [x] Add Authentication
- [ ] Add report ingestor
- [x] Add scan ingestors
- [x] Add Open API support
- [ ] File server
- [ ] Diagnosis

PR #969, opened 2023-03-25
Refactor probe report

ramanan-ravi:
#725


deepfence_server/ingesters/agent.go
```diff
@@ -237,168 +238,139 @@ func concatMaps(input map[string][]string) []map[string]interface{} {
...
-						}
-						right_ippid, ok := resolvers.get_ip_pid(ip + port)
-						if ok {
-							rightpid := extractPidFromNodeID(right_ippid)
+		if n.Pid >= 0 {
+			adjacency, ok := rpt.EndpointAdjacency[n.NodeID]
+			if !ok {
+				continue
+			}
+			if len(adjacency) == 0 {
```
noboruma, line 297: Not sure we need this condition?
I don't see it in prev code

deepfence_server/ingesters/agent.go
```diff
@@ -237,168 +238,139 @@ func concatMaps(input map[string][]string) []map[string]interface{} {
...
-		}
-		//node_info["adja"] = strings.Join(n.Adjacency, ",")
-		//endpoint_batch = append(endpoint_batch, node_info)
-
-		host_name := extractHostFromHostNodeID(node_info["host_node_id"])
-		if len(node_info["pid"]) > 0 && len(n.Adjacency) == 0 {
-			edges = append(edges,
-				map[string]string{"destination": host_name, "left_pid": "0", "right_pid": node_info["pid"]})
-		}
+	if len(inbound_edges) > 0 {
```
noboruma, line 325: We can remove that check, even if `inbound_edges` is empty, we expect to have a `in-the-internet` entry later

deepfence_server/ingesters/agent.go
```diff
@@ -237,168 +238,139 @@ func concatMaps(input map[string][]string) []map[string]interface{} {
...
+		endpoint_edges_batch = append(endpoint_edges_batch, map[string]interface{}{"source": "in-the-internet", "edges": inbound_edges})
 	}
-	endpoint_edges_batch = append(endpoint_edges_batch, map[string]interface{}{"source": "in-the-internet", "edges": edges})
 
-	process_batch := make([]map[string]string, 0, len(rpt.Process.Nodes))
+	process_batch := make([]map[string]interface{}, 0, len(rpt.Process))
 	process_edges_batch := map[string][]string{}
-	for _, n := range rpt.Process.Nodes {
-		node_info := n.ToDataMap()
-		host := node_info["node_id"]
```
noboruma, line 345: What is the value of `NodeID` here?
When constructing the graph, we expect the `node_id` of process to be in a specific format: `host_name;pid` to reconstruct connections quickly

deepfence_server/ingesters/agent.go
ramanan-ravi: Ok

deepfence_server/ingesters/agent.go
ramanan-ravi: There were 2 for loops, merged to one, so this condition. (mostly to avoid few lines of duplicate code)
https://github.com/deepfence/ThreatMapper/blob/v2/deepfence_server/ingesters/agent.go#L273-L340

deepfence_server/ingesters/agent.go
```diff
@@ -237,168 +238,139 @@ func concatMaps(input map[string][]string) []map[string]interface{} {
...
-				if n.ID != i {
-					ip, port := extractIPPortFromEndpointID(i)
-					if host, ok := resolvers.get_host(ip); ok {
-						if host_name == host {
-							continue
-						}
-						right_ippid, ok := resolvers.get_ip_pid(ip + port)
-						if ok {
-							rightpid := extractPidFromNodeID(right_ippid)
+		if n.Pid >= 0 {
```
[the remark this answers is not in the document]
ramanan-ravi: Ok

deepfence_server/ingesters/agent.go
ramanan-ravi: Process node_id format unchanged https://github.com/deepfence/ThreatMapper/blob/a53772ad0618b72dc8cbd25f7516808abd1bad15/deepfence_agent/tools/apache/scope/probe/process/reporter.go#L123

deepfence_server/ingesters/agent.go
noboruma: Ah I see - I personally like it separated because 1 code section is about linking "internal" connections, while the other code section is about linking "external" connections. But I get your point regarding code dup.
Can we just put back the comment: `// Handle inbound from internet` to indicate `len(adjacency) == 0` meaning more clearly

PR #977, opened 2023-03-29
Fix docker command and images tags

noriban:
Fixes # .
docker compose command in readme and images tags

@deepfence/engineering

when i turn on docker compose then i get error with this images


ramanan-ravi:
Thanks for the fix @noriban

PR #979, opened 2023-03-29
Fix commands in docs on website

noriban:
There is no more `docker-compose` command.

@deepfence/engineering


ramanan-ravi:
Thanks for the fix @noriban

PR #1025, opened 2023-04-19
v2-ui-settings-ui-improvements

harishshg:
Fixes # .

- changed user management as default landing tab
- reduced size forms in settings page to contain only width if the form
- added copy icon to resource column in user audit logs

@deepfence/engineering


milan-deepfence:
Invite user
Email field and role field looks different in sizes, not sure why.

Please check to add some top margin before button on those forms.
Also lets make sure that button are of same sizes across forms

manV (approved):

PR #1039, opened 2023-04-21
[UI][V2] Cloud Resources Details

manV:
- cloud resources details panel
- other improvements


manV:
there are some places where we handle aws only, we will add support for other clouds when api support lands

milan-deepfence (approved):

PR #1082, opened 2023-05-12
(feat.) add gitlab registry support

ibreakthecloud:
- adds support for gitlab as a registry

@deepfence/engineering

### DIFF
--- a/deepfence_worker/utils/registry.go
+++ b/deepfence_worker/utils/registry.go
@@ -17,6 +17,7 @@ import (
 	"github.com/deepfence/ThreatMapper/deepfence_server/pkg/registry/dockerprivate"
 	registryECR "github.com/deepfence/ThreatMapper/deepfence_server/pkg/registry/ecr"
 	"github.com/deepfence/ThreatMapper/deepfence_server/pkg/registry/gcr"
+	"github.com/deepfence/ThreatMapper/deepfence_server/pkg/registry/gitlab"
 	"github.com/deepfence/ThreatMapper/deepfence_server/pkg/registry/harbor"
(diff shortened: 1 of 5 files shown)


deepfence_worker/utils/registry.go
```diff
@@ -108,11 +109,52 @@ func GetCredentialsFromRegistry(ctx context.Context, registryId string) (regCred
...
+	if err != nil {
+		log.Error().Msg(err.Error())
+	}
+
+	return regCreds{
+		URL:           hub.NonSecret.GitlabRegistryURL,
+		UserName:      "gitlab-ci-token",
+		Password:      hub.Secret.GitlabToken,
+		NameSpace:     "",
+		ImagePrefix:   "registry.gitlab.com",
```
gnmahanth, line 152: image prefix is hard coded, what about the cases where its self hosted ?

deepfence_worker/utils/registry.go
```diff
@@ -108,11 +109,52 @@ func GetCredentialsFromRegistry(ctx context.Context, registryId string) (regCred
...
+		SkipTLSVerify: false,
```
gnmahanth, line 153: this needs to be true in case of self signed certificate, required as long as we dont implement a way to upload self-signed certificates

ibreakthecloud: currently we are not supporting self signed certificates, but will change once we have that feature.

deepfence_worker/utils/registry.go
ibreakthecloud: updated

PR #1092, opened 2023-05-17
(v2) trigger registry sync task when new registry is added

ibreakthecloud:
Triggers a registry sync task in deepfence_worker to sync single registry whenever a new registry is added.


gnmahanth (approved):
ok to merge

PR #1142, opened 2023-05-25
display toast message when count of nodes reach threshold value

milan-deepfence:
Set maximum number of nodes to render to 2k and display message when threshold(2k) is reached

@deepfence/engineering


deepfence_frontend/apps/dashboard/src/features/topology/components/TopologyGraph.tsx
```diff
@@ -31,6 +32,8 @@ import {
   GraphStorageManager,
 } from '@/features/topology/utils/topology-data';
 
+const MAX_NODES_COUNT_THRESHOLD = 10;
```
milan-deepfence, line 35: miss to set 2000

PR #1177, opened 2023-05-30
Fix scale & neo4j

noboruma:
Tweak numbers, add caching to address scalability high load
Optimize neo4j queries to make them more robust (add timeout, add more limits)

### DIFF
--- a/deepfence_worker/cronjobs/threat.go
+++ b/deepfence_worker/cronjobs/threat.go
@@ -2,10 +2,12 @@ package cronjobs
 
 import (
 	"sync/atomic"
+	"time"
 
 	"github.com/ThreeDotsLabs/watermill/message"
 
 	"github.com/deepfence/golang_deepfence_sdk/utils/directory"
+	"github.com/deepfence/golang_deepfence_sdk/utils/log"
 	"github.com/neo4j/neo4j-go-driver/v4/neo4j"
 )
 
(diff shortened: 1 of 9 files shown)


deepfence_worker/cronjobs/threat.go
```diff
@@ -2,14 +2,17 @@ package cronjobs
...
 
 	"github.com/ThreeDotsLabs/watermill/message"
 
 	"github.com/deepfence/golang_deepfence_sdk/utils/directory"
+	"github.com/deepfence/golang_deepfence_sdk/utils/log"
 	"github.com/neo4j/neo4j-go-driver/v4/neo4j"
 )
 
 var threatGraphRunning atomic.Bool
+var exploitabilityRunning atomic.Bool
```
gnmahanth, line 15: I think these flags needs to be set and checked from redis as the jobs might be running on different workers

noboruma: Good point, more globally I am not sure worker is actually a scalable process
For one, it has all the cron jobs, which should all be unique

noboruma:
Merged

deepfence_worker/cronjobs/threat.go
ramanan-ravi: Scheduler is unique and triggers cron jobs. Worker should be scalable. @noboruma

PR #1184, opened 2023-05-31
v2-ui-1175- Not Able to download Agent Diagnostic Logs

harishshg:
Fixes # .

- added active flag for hosts and Kubernetes clusters  selectors 

@deepfence/engineering

### DIFF
--- a/deepfence_frontend/apps/dashboard/src/features/common/data-component/searchClustersApiLoader.tsx
+++ b/deepfence_frontend/apps/dashboard/src/features/common/data-component/searchClustersApiLoader.tsx
@@ -35,7 +35,10 @@ export const searchClustersApiLoader = async ({
         filters: {
           compare_filter: null,
           contains_filter: {
-            filter_in: null,
+            filter_in: {
+              pseudo: [false],
+              active: [true],
+            },
           },
           match_filter: matchFilter,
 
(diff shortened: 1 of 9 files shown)


deepfence_frontend/apps/dashboard/src/features/common/data-component/searchClustersApiLoader.tsx
```diff
@@ -35,7 +35,10 @@ export const searchClustersApiLoader = async ({
         filters: {
           compare_filter: null,
           contains_filter: {
-            filter_in: null,
+            filter_in: {
+              pseudo: [false],
+              active: [true],
```
manV, line 40: It would make sense to show inactive nodes in the filters for example in the scan history page, we should make this configurable. Also if we are adding this here, we also need to add it for container and container images filter dropdowns.

manV: Can you please check with backend team and make sure?

ramanan-ravi: @manV @harishshg `active: [true]` is required in diagnosis page.
There are probably other cases where we need it.
There are also case where we don't need it.
So it has to be configurable.

ramanan-ravi: `pseudo: [false]` is required in all cases.

milan-deepfence: additional filters for all searches can be made as configurable

harishshg: added active filter flag at searchList components level for hosts, clustors, images and imageContainers.

manV (approved):

PR #1196, opened 2023-06-02
Use kafka for cloud resource ingestion

noboruma:


deepfence_worker/ingesters/cloud_resource.go
```diff
@@ -0,0 +1,348 @@
...
+		if len(splits) > 2 {
+			bb["cloud_provider"] = splits[2]
+		}
+	}
+
+	return bb
+}
+
+// TODO: Call somewhere
+func LinkNodesWithCloudResources(ctx context.Context) error {
```
gnmahanth, line 273: can you can call this at the end of commit ? what will be the impact ? after every commit this has to check and run on large number of data ?

do we need to add some kind of afterfunction for our bulk processor ?

gnmahanth (approved):

PR #1208, opened 2023-06-07
UI v2 new theme lables badges

harishshg:
Fixes # .

- added badges and labels

@deepfence/engineering


manV (changes requested):
Can you recheck this PR? in my quick glance I found that the colors, font sizes, line heights spacing deviating from the figma. we need to have it pixel perfect exactly same as the designs.

Also for colors, always use `dark:` prefix in the className, like we discussed.

manV:
there are some issue with badge, the height is not same as designs, text is not in the middle
image
image

manV:
Also double check paddings, we probably can't match figma 100% but the element at least should be same height width and it should look right.

manV:
the component variants can be improved. Ideally I would like two variants. 1. outlined and 2. filled. Two sizes 1. default/Regular 2. small.  That would cover all the badges. We can remove value prop since we are not going to support the mixed badge which is in the design.
image

manV:
Is this ready for review? @harishshg

harishshg:
updated the variant types

milan-deepfence (approved):

PR #1233, opened 2023-06-14
Upgrade golang version, dependencies, simplify agent build

ramanan-ravi:


deepfence_agent/Dockerfile
```diff
@@ -75,14 +75,15 @@ RUN apt-get update --allow-insecure-repositories \
     && cd /tmp \
     && chmod +x /usr/local/bin/start_services \
     && apt-get clean && apt-get -y autoremove && rm -rf /var/lib/apt/lists/* \
-    && curl -fsSLOk https://github.com/deepfence/vessel/releases/download/v0.9.0/vessel_v0.9.0_linux_amd64.tar.gz \
-    && tar -xzf vessel_v0.9.0_linux_amd64.tar.gz \
+    && curl -fsSLOk https://github.com/deepfence/vessel/releases/download/v0.11.0/vessel_v0.11.0_linux_amd64.tar.gz \
```
shyam-dev, line 78: Can we replace this with a command to download the current latest version of vessel ? If not, whenever we release an update to vessel, we will have to change this line in code and update agent too. @noboruma one more for the plugin I suppose ?

deepfence_agent/tools/apache/deepfence/df-utils/go.sum
```diff
@@ -159,8 +159,8 @@ github.com/shurcooL/sanitized_anchor_name v0.0.0-20150822220530-244f5ac324cb/go.
 github.com/sirupsen/logrus v1.0.6/go.mod h1:pMByvHTf9Beacp5x1UXfOR9xyW/9antXMhjMPG0dEzc=
 github.com/sirupsen/logrus v1.2.0/go.mod h1:LxeOpSwHxABJmUn/MG1IvRgCAasNZTLOkJPxbbu5VWo=
 github.com/sirupsen/logrus v1.4.2/go.mod h1:tLMulIdttU9McNUspp0xgXVQah82FyeX6MwdIuYE2rE=
-github.com/sirupsen/logrus v1.9.0 h1:trlNQbNUG3OdDrDil03MCb1H2o9nJ1x4/5LYw7byDE0=
-github.com/sirupsen/logrus v1.9.0/go.mod h1:naHLuLoDiP4jHNo9R0sCBMtWGeIprob74mVsIT4qYEQ=
+github.com/sirupsen/logrus v1.9.3 h1:dueUQJ1C2q9oE3F7wvmSGAaVtTmUizReu6fjN8uqzbQ=
```
shyam-dev, line 162: can we add a gitignore so that go.sum does not get checked in but go.mod does ?

ramanan-ravi: https://stackoverflow.com/a/65205889/2915763

deepfence_agent/Dockerfile
ramanan-ravi: I would not want latest vessel version to be downloaded automatically.
I need to use v0.10.0 in TM v1.5 and v0.11.0 in TM v2.
(Latest vessel may not work with TM v1.5 because golang version is different)

shyam-dev: go.mod is a potential solution path

noboruma: We need to agree on our update strategy - I agree with Ramanan, we don't want any automatic upgrade as this might affect users, we need to ask them first and then proceed.
With that being said, we definitely need a way to upgrade vessel. Here we have two ways:
- Either we do a generic upgrade: all binaries are bundled together and we extract them once together
- Or we do partial upgrades for each individual components/plugins: binaries are upgrade separately

Both have advantages - problem with second approach is we need to make sure every piece of up-gradable code is a plugin, which is not the case here with vessel (it's a utils tool)

deepfence_agent/Makefile
```diff
@@ -38,6 +40,17 @@ gocode:
 
 install:
 
+agent:
```
noboruma, line 43: At last! Great we finally have it :-)

gnmahanth (approved):

deepfence_agent/Makefile
```diff
@@ -38,6 +40,17 @@ gocode:
...
+	$(CD) tools/apache/scope && \
+		go mod tidy -v && \
+		go mod vendor && \
+		env GOGC=off \
+		go build \
+		-ldflags "-X main.version=$(shell git rev-parse --short HEAD)  -X github.com/weaveworks/scope/probe/host.agentCommitID=$(shell git rev-parse --short HEAD) -X github.com/weaveworks/scope/probe/host.agentBuildTime=$(shell date +"%s%d%m%y") -s -w" \
+		-tags 'netgo osusergo unsafe' \
+		-o docker/deepfence_exe \
+		./prog
```
noboruma, line 52:
```suggestion
	($(CD) tools/apache/scope && \
		go mod tidy -v && \
		go mod vendor && \
		env GOGC=off \
		go build \
		-ldflags "-X main.version=$(shell git rev-parse --short HEAD)  -X github.com/weaveworks/scope/probe/host.agentCommitID=$(shell git rev-parse --short HEAD) -X github.com/weaveworks/scope/probe/host.agentBuildTime=$(shell date +"%s%d%m%y") -s -w" \
		-tags 'netgo osusergo unsafe' \
		-o docker/deepfence_exe \
		./prog)
```
This will prevent the shell from changing its location after doing `make agent`

noboruma (approved):

PR #1239, opened 2023-06-19
Discover cloud compute instances from cloud scanner in agent-less deployment

ramanan-ravi:
https://github.com/deepfence/enterprise-roadmap/issues/1811


deepfence_worker/cronjobs/neo4j.go
```diff
@@ -100,7 +100,8 @@ func CleanUpDB(msg *message.Message) error {
 	if _, err = session.Run(`
 		MATCH (n:Node)
 		WHERE n.updated_at < TIMESTAMP()-$time_ms
-		AND NOT n.node_id IN ["in-the-internet", "out-the-internet"]
+		AND NOT n.node_id IN ["in-the-internet", "out-the-internet"] 
+		AND n.agent_running=true
```
noboruma, line 104: `agent_running` is deprecated, it is not used anywhere.
Its replacement is `active`.
We should remove this entirely from the agent itself

deepfence_worker/cronjobs/neo4j.go
```diff
@@ -109,6 +110,18 @@ func CleanUpDB(msg *message.Message) error {
 		return err
 	}
 
+	if _, err = session.Run(`
+		MATCH (n:Node)
+		WHERE n.updated_at < TIMESTAMP()-$time_ms
+		AND n.agent_running=false
```
noboruma, line 116: This should be `agent_running=true` here
`agent_running` now means that an agent is deployed, `active` should be used to determine is an agent is active/running.
Maybe we should rename `agent_running` to `agent_present` to avoid such mistakes?

deepfence_worker/ingesters/cloud_resource.go
```diff
@@ -131,6 +144,24 @@ func CommitFuncCloudResource(ns string, cs []CloudResource) error {
...
+		MERGE (m:Node{node_id:row.node_id})
+		MERGE (cp:CloudProvider{node_id:row.cloud_provider})
+		MERGE (cr:CloudRegion{node_id:row.cloud_region})
+		MERGE (cp) -[:HOSTS]-> (cr)
+		MERGE (cr) -[:HOSTS]-> (m)
+		SET m+=row, m.updated_at = TIMESTAMP(), cp.active = true, cp.pseudo = false, cr.active = true`,
+			map[string]interface{}{"batch": hosts},
+		)
+	}
+
```
noboruma, line 164: Maybe not for this PR but we need to move this piece of code here instead:
https://github.com/deepfence/ThreatMapper/blob/v2/deepfence_worker/cronjobs/neo4j.go:356

This is purely for performance reason, accessing `CloudProvider` and `CloudRegion` will result in deadlock situations under heavy load

noboruma (approved):
LGTM

PR #1310, opened 2023-07-11
UI v2 new theme feedback 1

milan-deepfence:


deepfence_frontend/packages/ui-components/src/components/tooltip/Tooltip.tsx
```diff
@@ -26,7 +26,7 @@ export const Tooltip = (props: TooltipProps) => {
     label,
   } = props;
   return (
-    <TooltipPrimitive.Provider delayDuration={delayDuration ?? 0}>
+    <TooltipPrimitive.Provider delayDuration={delayDuration ?? 2}>
```
manV, line 29: This does not solve the issue. The problem is the tooltip on <TruncatedText> component is controlled one and delayDuration has no affect on controlled tooltip.

manV (approved):

PR #1316, opened 2023-07-12
ISSUE-1830: Adding changes for stopping the seceret and malware scans (Only review, do not merge yet)

varunsharma0286:
Fixes #
https://github.com/deepfence/enterprise-roadmap/issues/1830
- Adding support for stopping the secret scans 
- Adding support for stopping the malware scans

### DIFF
--- a/deepfence_bootstrapper/router/secret_scanner.go
+++ b/deepfence_bootstrapper/router/secret_scanner.go
@@ -111,3 +111,26 @@ func GetSecretScannerJobCount() int32 {
 	}
 	return jobReport.RunningJobs
 }
+
+func StopSecretScan(req ctl.StopSecretScanRequest) error {
+	fmt.Printf("Stop Secret Scan : %v\n", req)
+	conn, err := grpc.Dial("unix://"+ebpfSocketPath, grpc.WithAuthority("dummy"),
+		grpc.WithInsecure())
+	if err != nil {
+		fmt.Printf("error in creating secret scanner client: %s\n", err.Error())
+		return err
+	}
+
+	defer conn.Close()
+	client := pb.NewScannersClient(conn)
+
+	var greq pb.StopScanRequest
+	greq.ScanId = req.BinArgs["scan_id"]
+
+	_, err = client.StopScan(context.Background(), &greq)
+	if err != nil {
+		fmt.Printf("StopSecretScan::error in client.StopScan: %s\n", err.Error())
+	}
+
+	return err
+}
--- a/deepfence_ctl/cmd/scan.go
+++ b/deepfence_ctl/cmd/scan.go
@@ -397,6 +397,51 @@ var scanResultsSubCmd = &cobra.Command{
 	},
 }
 
+var scanStopSubCmd = &cobra.Command{
+	Use:   "stop",
+	Short: "Stop Scan",
+	Long:  `This subcommand stops a scan`,
+	Run: func(cmd *cobra.Command, args []string) {
+
+		log.Info().Msgf("Command: %v", *cmd)
+		scan_type, _ := cmd.Flags().GetString("type")
+		if scan_type == "" {
+			log.Fatal().Msg("Please provide an type")
+		}
+
+		scan_id, _ := cmd.Flags().GetString("scan-id")
+		if scan_id == "" {
+			log.Fatal().Msg("Please provide a scan id")
+		}
+
+		var err error
+		var res interface{}
+		switch scan_type {
+		case "secret":
+			req := http.Client().SecretScanAPI.StopSecretScan(context.Background())
+			req = req.ModelStopScanRequest(deepfence_server_client.ModelStopScanRequest{
+				ScanId:   scan_id,
+				ScanType: "SecretScan",
+			})
+			res, err = http.Client().SecretScanAPI.StopSecretScanExecute(req)
+		case "malware":
+			req := http.Client().MalwareScanAPI.StopMalwareScan(context.Background())
+			req = req.ModelStopScanRequest(deepfence_server_client.ModelStopScanRequest{
+				ScanId:   scan_id,
+				ScanType: "MalwareScan",
+			})
+			res, err = http.Client().MalwareScanAPI.StopMalwareScanExecute(req)
+		default:
+			log.Fatal().Msg("Unsupported")
+		}
+
+		if err != nil {
+			log.Fatal().Msgf("Fail to execute: %v", err)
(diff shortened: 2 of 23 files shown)


gnmahanth (approved):

deepfence_bootstrapper/router/secret_scanner.go
```diff
@@ -111,3 +111,26 @@ func GetSecretScannerJobCount() int32 {
 	}
 	return jobReport.RunningJobs
 }
+
+func StopSecretScan(req ctl.StopSecretScanRequest) error {
+	fmt.Printf("Stop Secret Scan : %v\n", req)
+	conn, err := grpc.Dial("unix://"+ebpfSocketPath, grpc.WithAuthority("dummy"),
+		grpc.WithInsecure())
```
noboruma, line 118: Not for this PR but we probably need to revamp so that we only have 1 client.
With that said, maybe gRPC client is already clever enough and reuse connections

deepfence_ctl/cmd/scan.go
```diff
@@ -397,13 +397,59 @@ var scanResultsSubCmd = &cobra.Command{
 	},
 }
 
+var scanStopSubCmd = &cobra.Command{
```
noboruma, line 400: Nice to see `deepfence_ctl` getting some love :)

deepfence_server/controls/agent.go
```diff
@@ -14,22 +14,37 @@ import (
 )
 
 var (
+	MaxStops    = 5
```
noboruma, line 17: This `MaxStops` is a bit misleading because it only applies for 1 iteration: Let's imagine we cancel 30 scans, we will do: 5 cancel, next 30 seconds 5 more, next 30 seconds 5 more, without checking if the previous cancel were completed.

Now I think in practice we should be fine because we expect cancels to finish quickly and the request should not use extra resources to complete.. I would rather remove that `MaxStops` entirely, just to simplify our logic and not mislead us in the future

deepfence_utils/utils/constants.go
```diff
@@ -47,13 +47,18 @@ const (
...
 const (
-	SCAN_STATUS_SUCCESS    = "COMPLETE"
-	SCAN_STATUS_STARTING   = "STARTING"
-	SCAN_STATUS_INPROGRESS = "IN_PROGRESS"
-	SCAN_STATUS_FAILED     = "ERROR"
+	SCAN_STATUS_SUCCESS        = "COMPLETE"
+	SCAN_STATUS_STARTING       = "STARTING"
+	SCAN_STATUS_INPROGRESS     = "IN_PROGRESS"
+	SCAN_STATUS_FAILED         = "ERROR"
+	SCAN_STATUS_CANCELLING_REQ = "CANCELLING_REQ"
```
noboruma, line 59: I would rename this either:
- `SCAN_STATUS_CANCEL_REQ = CANCEL_REQ` 
- `SCAN_STATUS_CANCEL_PENDING = CANCEL_PENDING`
With a slight preference for the second one

deepfence_worker/ingesters/common.go
```diff
@@ -88,7 +88,16 @@ func CommitFuncStatus[Status any](ts utils.Neo4jScanType) func(ns string, data [
...
 
-		if _, err = tx.Run(query, map[string]interface{}{"batch": statusesToMaps(data)}); err != nil {
+		maps := statusesToMaps(data)
+		for idx := 0; idx < 3; idx++ {
+			if _, err = tx.Run(query, map[string]interface{}{"batch": maps}); err != nil {
+				log.Error().Msgf("Error occured while updating scan status: %+v, Retrying...", err)
+				time.Sleep(1 * time.Second)
+				//return err
+			}
+		}
```
noboruma, line 98: This looks like a retry loop:
- rename `idx` into `retries`
- add a `break` in the `else` branch

noboruma: Actually none of this is required
If an ingestion has an error, the upper function will take care of retry

deepfence_worker/ingesters/common.go
```diff
@@ -97,31 +106,41 @@ func CommitFuncStatus[Status any](ts utils.Neo4jScanType) func(ns string, data [
...
 
-// also handles status deduplication
+// handles status deduplication
+// and maintains the order of the messages
 func statusesToMaps[T any](data []T) []map[string]interface{} {
-	statusBuff := map[string]map[string]interface{}{}
+	type Entry struct {
+		Posn int
+		Data map[string]interface{}
+	}
```
noboruma, line 115: I would rather keep previous code here as this is not solving anything new from what I understand.
I am fine with the allocation optimization:
```
statuses := make([]map[string]interface{}, len(statusBuff))
```
But the `Posn` introduction is not solving anything new, is it?

deepfence_server/controls/agent.go
varunsharma0286: I believe we would need to put in a limit to how many we can process in a one go.   
Maybe we can raise this to higher number (maybe 50)

deepfence_server/handler/scan_reports.go
```diff
@@ -674,7 +674,72 @@ func ingest_scan_report_kafka[T any](
 }
 
 func stopScan(w http.ResponseWriter, r *http.Request, action ctl.ActionID) {
+
+}
+
+func (h *Handler) stopSecretScan(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
varunsharma0286: Will make this change for 2nd PR.

deepfence_utils/utils/constants.go
varunsharma0286: I will rename this to "CANCEL_PENDING"

deepfence_worker/ingesters/common.go
varunsharma0286: Done.

deepfence_utils/utils/constants.go
varunsharma0286: Done.

deepfence_worker/ingesters/common.go
```diff
@@ -130,6 +149,7 @@ func ToMap[T any](data T) map[string]interface{} {
 	if err != nil {
 		return nil
 	}
+	//log.Info().Msg("VARUN:: ToMap -> " + string(out))
```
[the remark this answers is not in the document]
varunsharma0286: Done.

PR #1320, opened 2023-07-13
reduce CloudNode cleanup time from 13 hours to 5 mins

ibreakthecloud:


ibreakthecloud:
> I am not sure about that one - if nodes are only updated every 12h, 5 minutes will be too short and start removing nodes early, no?

I think the nodes are updated every minute
https://github.com/deepfence/cloud-scanner/blob/v2/service/service.go#L124

ramanan-ravi:
Cloud registration happens every minute. Cloud resources are ingested once a day.

PR #1378, opened 2023-07-19
Fix GCR get_images_list

p-hash:
Currently deepfence does not list any images when configured to scan GCR (Google Container Registry).

This happens because `https://gcr.io/v2/_catalog` always returns `{"repositories":[]}`, even with valid credentials to private repository.

To enumerate repositories, we should call `https://gcr.io/v2/<project_id>/tags/list` and iterate over `"child":[]` list. Each of those children could be image or repository of images with its own children. So the solution here is a recursive search down the tree.


shyam-dev:
Thank you for the submission, @p-hash !

ramanan-ravi (approved):

PR #1386, opened 2023-07-20
(v2) donot allow email integration if not configured

ibreakthecloud:
- refactor integration to accept `context` in `New()`
- donot allow user to add email in integration if email os not configured

### DIFF
--- a/deepfence_server/handler/integration.go
+++ b/deepfence_server/handler/integration.go
@@ -26,6 +26,13 @@ func (h *Handler) AddIntegration(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	ctx := r.Context()
+	pgClient, err := directory.PostgresClient(ctx)
+	if err != nil {
+		respondError(&InternalServerError{err}, w)
+		return
+	}
+
 	// identify integration and interface it
 	b, err := json.Marshal(req)
 	if err != nil {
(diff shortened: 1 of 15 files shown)


ibreakthecloud:
todo: refactor: avoid passing `context` in integration methods. Set in the the integration object and use it there to avoid context duplication

ramanan-ravi (approved):

ibreakthecloud:
> @ibreakthecloud Should we warn user when email configuration is deleted and there is an active email integration?

makes sense.
So, 
- should we restrict user from deleting Email config, and tell them to delete email integrations first if email integration exists or, 
- should we show generic message while deleting saying "All you email integration will be deleted", irrespective of email integration exists and if they continue we delete email integrations too?

ramanan-ravi:
I think we should just show a warning message. (May be a double confirmation popup like "do you still want to proceed?")

We should not delete integrations. There was time involved in configuring them. Asking to reconfigure is not good experience.

A genuine case is SES keys changed. I delete and create. Or I move from SMTP to SES.

shyam-dev:
Valid point. So.

While deleting an email configuration, if email integrations are configured, we just show a message that email integrations will not work anymore.

ibreakthecloud:
> Valid point. So.
> 
> While deleting an email configuration, if email integrations are configured, we just show a message that email integrations will not work anymore.

This becomes a separate ticket, since the changes are only in frontend, will create.

ibreakthecloud:
Merging this for now on the @ramanan-ravi approval
thanks

Issue #1332, opened 2023-07-17
[v2] No scan status when trying to start scan


saurabh2253:
**Describe the bug**
When trying to start a scan for vulnerability there is no check if the subject is already under scan. One can start multiple scans on same container while a scan on it is already in progress.

<img width="522" alt="Screenshot 2023-07-17 at 6 56 32 PM" src="https://github.com/deepfence/ThreatMapper/assets/3922581/54351fbc-f2d3-4132-9b3f-15e65751d2e0">
<img width="641" alt="Screenshot 2023-07-17 at 6 53 42 PM" src="https://github.com/deepfence/ThreatMapper/assets/3922581/90de9f3e-7546-4c5d-a096-d6ad095ac37c">
<img width="651" alt="Screenshot 2023-07-17 at 6 53 27 PM" src="https://github.com/deepfence/ThreatMapper/assets/3922581/ac9f5df2-90ac-45a3-98ca-a5820fd846d9">

**To Reproduce**
1. Select any container from topology  and open side panel
2. Start Vuln Scan
3. Start Vuln scan again immediately

**Expected behavior**
A container with scan in progress should be available for scan once in progress scan is complete.

gnmahanth:
These are the recommended steps for the UI, on the side-panel or in the table view 
1. Gray out the option to start new scans on the nodes where scans are already `in progress` or `starting` so that that it indicates no new scan can be started when the scans are running 
2. if UI sees some of the scans were already in progress when a bulk scan is started, show a message saying some of the nodes were skipped since scan is running 

@ramanan-ravi @noboruma @shyam-dev @manV

noboruma:
NB: The above recommendation is taking into account the fact UI might not be updated to the latest state of scan statuses. First, doing a scan back-to-back is a very unlikely real-life scenario, and if someone wants it, they can hit the refresh button.
Secondly, while UI should have as few logic as possible, we should avoid unnecessary calls to the backend - by greying out a button & or opting out some ID, we are making the UX clearer & avoid unnecessary bandwidth usage

manV:
I am not sure if we should grey out the buttons. What if a scan is can be started but the ui is having a stale value status in progress.

I would suggest let the users start the scan regardless of the current status. If scan cannot be started, we show the error message. In real world scenario this is unlikely like Thomas mentioned, why add so much logic of enabling/disabling buttons etc? In this case a couple of extra API calls would hardly be a problem.

shyam-dev:
[...]

PR #1392, opened 2023-07-21
Fix: Check and skip starting scans if a scan is already running

gnmahanth:
- https://github.com/deepfence/ThreatMapper/issues/1332

### DIFF
--- a/deepfence_server/handler/scan_reports.go
+++ b/deepfence_server/handler/scan_reports.go
@@ -229,7 +229,7 @@ func (h *Handler) StartVulnerabilityScanHandler(w http.ResponseWriter, r *http.R
 
 	actionBuilder := StartScanActionBuilder(r.Context(), ctl.StartVulnerabilityScan, binArgs)
 
-	scan_ids, bulkId, err := StartMultiScan(r.Context(), true, utils.NEO4J_VULNERABILITY_SCAN, reqs.ScanTriggerCommon, actionBuilder)
+	scan_ids, bulkId, skipped, err := StartMultiScan(r.Context(), true, utils.NEO4J_VULNERABILITY_SCAN, reqs.ScanTriggerCommon, actionBuilder)
 	if err != nil {
 		if err.Error() == "Result contains no more records" {
 			respondError(&noNodesMatchedInNeo4jError, w)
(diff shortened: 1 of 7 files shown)


deepfence_server/handler/scan_reports.go
```diff
@@ -1877,23 +1922,32 @@ func StartMultiScan(ctx context.Context,
...
+			return nil, "", skip_node_ids, err
 		}
 		reqs = append(reqs, reqs_extra...)
 	} else {
 		reqs = req.NodeIds
 	}
 
+	// check and skip if scan are already running or scheduled
+	for _, n := range reqs {
+		if IsScanInProgress(ctx, scan_type, n) {
```
ramanan-ravi, line 1934: I think it will be better to get all currently running scans in one query and compare it against the map.

gnmahanth: If I want to query all node_ids it will be a where query and might take longer to complete, this is a simple read query should be faster and ok to run multiple times, also as of now we can start scans on 50nodes at at time so this should be ok

deepfence_server/handler/scan_reports.go
```diff
@@ -1822,49 +1820,96 @@ func extractBulksNodes(nodes []model.NodeIdentifier) (regularNodes []model.NodeI
...
+	val, found := record.Get("status")
+	if !found {
+		return false
+	}
+
+	status := val.(string)
+	if status == utils.SCAN_STATUS_STARTING || status == utils.SCAN_STATUS_INPROGRESS ||
+		status == utils.SCAN_STATUS_CANCELLING || status == utils.SCAN_STATUS_CANCEL_PENDING {
+		return true
+	}
```
noboruma, line 1863: This can be simplified with 1 neo4j query, you can take example on:
https://github.com/deepfence/ThreatMapper/blob/v2/deepfence_server/controls/version.go#L163

deepfence_server/handler/scan_reports.go
noboruma: That's a tricky one, I agree with Ramanan, having 1 read instead of N is preferable, we just need to replace:
```
Match (n:%s{node_id:$node_id})
```
with
```
MATCH (n:%s) WHERE n.node IN $all_ids
```

Now I am not a big fan of relying on a map in memory to determine if we should start a scan or not. We could handle that in the start scan neo4j query itself. It used to be done like that (and I am surprised to see this was removed over time, we should have it in the git history?)
The way it used to work is: we were starting all scans and if the scan was IN_PROGRESS, we were passing the old in_progress scan_id as if it was started.
It could be done via a combination of `OPTIONAL MATCH` and `MERGE`

gnmahanth:
these changes are not necessary the current behaviour is correct

Issue #1350, opened 2023-07-18
[v2] starting a scan on any single image tag triggers scan on all image tags for that image


manV:
**Describe the bug**
Registries - starting a scan on any single image tag triggers scan on all image tags for that image

**To Reproduce**
1. Go to any registry image tags page
2. start any scan on any single tag
3. scan is started for all tags for the same image.

**Expected behavior**
It should start scan on a specific tag

**Components/Services affected**

- [x] UI/Frontend
- [x] API/Backend
- [ ] Agent
- [ ] Deployment/YAMLs
- [ ] CI/CD Integration
- [ ] Other (specify)

milan-deepfence:
API request for start scan of image tag for instance:
{
  "filters": {
    "cloud_account_scan_filter": { "filter_in": null },
    "container_scan_filter": { "filter_in": null },
    "host_scan_filter": { "filter_in": null },
    "image_scan_filter": { "filter_in": null },
    "kubernetes_cluster_scan_filter": { "filter_in": null }
  },
  "node_ids": [
    {
      "node_id": "7fa35ea714050361090fa28ea09a1b4f234af8365ea79ac4925ec325662a1d8c",
      "node_type": "image"
    }
  ]
}

Where we see presence of a single node_id.
This could happen because of node ids are same for those tags:
[
    {
        "node_id": "7fa35ea714050361090fa28ea09a1b4f234af8365ea79ac4925ec325662a1d8c",
        "node_name": "docker-extension:0.0.2",
        "docker_image_name": "docker-extension",
        "docker_image_tag": "0.0.3",
        "docker_image_size": "1282",
        "docker_image_created_at": "",
        "docker_image_virtual_size": "",
        "docker_image_id": "7fa35ea714050361090fa28ea09a1b4f234af8365ea79ac4925ec325662a1d8c",
        "metadata": {
            "digest": "sha256:7fa35ea714050361090fa28ea09a1b4f234af8365ea79ac4925ec325662a1d8c",
            "last_pushed": 1679356130,
            "last_updated": 1679356131,
            "status": "inactive"
        },
        "vulnerabilities_count": 0,
        "vulnerability_scan_status": "ERROR",
        "vulnerability_latest_scan_id": "7fa35ea714050361090fa28ea09a1b4f234af8365ea79ac4925ec325662a1d8c-1689652124",
        "secrets_count": 0,
        "secret_scan_status": "STARTING",
[...]

PR #1398, opened 2023-07-24
ISSUE-1350: Fixing the docker registry

varunsharma0286:
Fixes #
https://github.com/deepfence/ThreatMapper/issues/1350

Also has the backend changes required for:
https://github.com/deepfence/enterprise-roadmap/issues/1901

- Add 2 new fields "docker_image_tag_list" and "docker_image_name_list" to the ContainerImage node.

### DIFF
--- a/deepfence_server/model/lookup.go
+++ b/deepfence_server/model/lookup.go
@@ -309,14 +309,18 @@ func (IngestedContainerImage) GetJsonCategory() string {
 }
 
 type ContainerImage struct {
-	ID                        string      `json:"node_id" required:"true"`
-	NodeName                  string      `json:"node_name" required:"true"`
-	Name                      string      `json:"docker_image_name" required:"true"`
+	ID       string `json:"node_id" required:"true"`
+	NodeName string `json:"node_name" required:"true"`
+	//Deprecated: Please use DockerImageNameList
+	Name string `json:"docker_image_name" required:"true"`
+	//Deprecated: Please use DockerImageTagList
 	Tag                       string      `json:"docker_image_tag" required:"true"`
 	Size                      string      `json:"docker_image_size" required:"true"`
 	DockerImageCreatedAt      string      `json:"docker_image_created_at" required:"true"`
 	DockerImageVirtualSize    string      `json:"docker_image_virtual_size" required:"true"`
 	DockerImageID             string      `json:"docker_image_id" required:"true"`
+	DockerImageTagList        []string    `json:"docker_image_tag_list" required:"true"`
+	DockerImageNameList       []string    `json:"docker_image_name_list" required:"true"`
 	Metadata                  Metadata    `json:"metadata" required:"true" nested_json:"true"`
 	VulnerabilitiesCount      int64       `json:"vulnerabilities_count" required:"true"`
 	VulnerabilityScanStatus   string      `json:"vulnerability_scan_status" required:"true"`
--- a/deepfence_server/model/registry.go
+++ b/deepfence_server/model/registry.go
@@ -397,7 +397,7 @@ func ListImages(ctx context.Context, registryId string, filter reporters.FieldsF
 			continue
 		}
 
-		l, ok := lValue.(dbtype.Node)
+		_, ok := lValue.(dbtype.Node)
 		if !ok {
 			log.Warn().Msgf("Missing neo4j entry")
 			continue
@@ -411,26 +411,7 @@ func ListImages(ctx context.Context, registryId string, filter reporters.FieldsF
 
 		var node ContainerImage
 		utils.FromMap(m.Props, &node)
-
-		// Extract tags from the "l" node
-		tagsValue, hasTags := l.Props["tags"]
-		if hasTags {
-			tags, ok := tagsValue.([]interface{})
-			if ok {
-				for _, tag := range tags {
-					if tagStr, ok := tag.(string); ok {
-						// create individual tag nodes
-						tagNode := node
-						tagNode.Tag = tagStr
-						res = append(res, tagNode)
-					}
-				}
-			}
-		}
-
-		// kludge: breaking here since we don't want to repeat images and duplicate the entries
-		// in the first iteration itself we get all the tags in an array and we use the first image and duplicate
-		break
+		res = append(res, node)
 	}
 
 	return res, nil
--- a/deepfence_utils/utils/utils.go
+++ b/deepfence_utils/utils/utils.go
@@ -263,8 +263,21 @@ func FromMap(bb map[string]interface{}, c interface{}) {
 		if t.Field(i).Type.Kind() == reflect.Slice {
 			slice, ok := data.([]map[string]interface{})
 			if !ok {
-				vv := reflect.ValueOf(data).Convert(t.Field(i).Type)
-				v.Field(i).Set(vv)
+				if t.Field(i).Type.Elem().Kind() == reflect.String {
+					//We are not able to convert the []interface{} to []string
+					//Hence we need to have this special handling
+					var outStr []string
+					rv := reflect.ValueOf(data)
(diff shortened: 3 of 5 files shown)


deepfence_server/model/registry.go
```diff
@@ -374,9 +374,9 @@ func ListImages(ctx context.Context, registryId string, filter reporters.FieldsF
 	}
 
 	query := `
-	MATCH (n:RegistryAccount{node_id: $id}) -[:HOSTS]-> (l:ImageStub) <-[:IS]- (m:ContainerImage)
-	` + reporters.ParseFieldFilters2CypherWhereConditions("l", mo.Some(filter), true) + `
-	RETURN l, m
```
noboruma, line 379: Why are we removing `:IS`?
I think this needs to stay since the `:IS` relationship allows extra data to be populated in the final object ended over by the rest API

deepfence_server/model/registry.go
```diff
@@ -390,15 +390,8 @@ func ListImages(ctx context.Context, registryId string, filter reporters.FieldsF
...
 	for _, rec := range records {
-		lValue, hasL := rec.Get("l")
 		mValue, hasM := rec.Get("m")
-		if !hasL || !hasM {
-			log.Warn().Msgf("Missing neo4j entry")
-			continue
-		}
-
-		l, ok := lValue.(dbtype.Node)
-		if !ok {
```
noboruma, line 401: Ditto, should stay

deepfence_server/model/lookup.go
```diff
@@ -309,25 +309,27 @@ func (IngestedContainerImage) GetJsonCategory() string {
...
+	ID                        string        `json:"node_id" required:"true"`
+	NodeName                  string        `json:"node_name" required:"true"`
+	Name                      string        `json:"docker_image_name" required:"true"`
+	Tag                       string        `json:"docker_image_tag" required:"true"`
+	Size                      string        `json:"docker_image_size" required:"true"`
+	DockerImageCreatedAt      string        `json:"docker_image_created_at" required:"true"`
+	DockerImageVirtualSize    string        `json:"docker_image_virtual_size" required:"true"`
+	DockerImageID             string        `json:"docker_image_id" required:"true"`
+	DockerImageTagList        []interface{} `json:"docker_image_tag_list" required:"true"`
+	DockerImageNameList       []interface{} `json:"docker_image_name_list" required:"true"`
```
noboruma, line 321: Can we have those as `[]string`?
Will make it nicer on OpenAPI

deepfence_server/model/lookup.go
```diff
@@ -309,25 +309,27 @@ func (IngestedContainerImage) GetJsonCategory() string {
...
-	SecretScanStatus          string      `json:"secret_scan_status" required:"true"`
-	SecretLatestScanId        string      `json:"secret_latest_scan_id" required:"true"`
-	MalwaresCount             int64       `json:"malwares_count" required:"true"`
-	MalwareScanStatus         string      `json:"malware_scan_status" required:"true"`
-	MalwareLatestScanId       string      `json:"malware_latest_scan_id" required:"true"`
-	Containers                []Container `json:"containers" required:"true"`
+	ID                        string        `json:"node_id" required:"true"`
+	NodeName                  string        `json:"node_name" required:"true"`
+	Name                      string        `json:"docker_image_name" required:"true"`
+	Tag                       string        `json:"docker_image_tag" required:"true"`
```
noboruma, line 315: We need to mark those as deprecated (new github issue + comment)

deepfence_server/model/registry.go
varunsharma0286: I am not sure why the ":IS" relationship is required in this case. 
Reason being, with the ":IS" query, we are getting the extra tags (the one which is retagged or duplicated from image of other repo). 
From what i see, we need to list all the images for the selected "repository" and "regsitry".  For that,  the above query should be sufficient (unless i am missing something).

deepfence_server/model/registry.go
```diff
@@ -412,25 +405,7 @@ func ListImages(ctx context.Context, registryId string, filter reporters.FieldsF
...
-						tagNode.Tag = tagStr
-						res = append(res, tagNode)
-					}
-				}
-			}
-		}
-
-		// kludge: breaking here since we don't want to repeat images and duplicate the entries
-		// in the first iteration itself we get all the tags in an array and we use the first image and duplicate
-		break
```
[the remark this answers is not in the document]
varunsharma0286: This is actually a bug. We are just using the first node from the result set "m" and updating the tag for that node. 
Also, since we don't have any way to associate result set "m" with tags in "l" , this would be incorrect. 
We would have to loop through the results in "m" and populate the result in "res".

deepfence_server/model/lookup.go
varunsharma0286: Earlier i used `[] string`. But this gave me a compilation error from the reflection. 
I will check if we can modify the code in the utils to support `[]string`

deepfence_server/model/registry.go
noboruma: Hum.. I see why you made that change now - I still think we need keep the `:IS` relationship in, to be consistent across the board: all nodes with `:IS` relationship are done this way.
This actually raises a bigger question whether `ImageStub` now makes sense or not. So I would rather either remove `ImageStub` (and the `:IS` binding) entirely, or keep it. The one use case it solves (not sure we use it) is to retrieve all tags associated with a given name, so it still brings some value

deepfence_server/model/lookup.go
noboruma: We are able to treat `[]Container` so I am surprised `[]string` is not supported

varunsharma0286: Pushed the change to handle the slice of strings.

deepfence_server/model/registry.go
```diff
@@ -397,7 +397,7 @@ func ListImages(ctx context.Context, registryId string, filter reporters.FieldsF
 			continue
 		}
 
-		l, ok := lValue.(dbtype.Node)
+		_, ok := lValue.(dbtype.Node)
```
noboruma, line 400:
```suggestion
		l, ok := lValue.(dbtype.Node)
```

deepfence_server/model/registry.go
```diff
@@ -411,26 +411,7 @@ func ListImages(ctx context.Context, registryId string, filter reporters.FieldsF
...
-						res = append(res, tagNode)
-					}
-				}
-			}
-		}
-
-		// kludge: breaking here since we don't want to repeat images and duplicate the entries
-		// in the first iteration itself we get all the tags in an array and we use the first image and duplicate
-		break
+		res = append(res, node)
```
noboruma, line 414:
```suggestion
        utils.FromMap(l.Props, &node)
		res = append(res, node)
```

deepfence_utils/utils/utils.go
```diff
@@ -263,8 +263,21 @@ func FromMap(bb map[string]interface{}, c interface{}) {
...
+					for i := 0; i < rv.Len(); i++ {
+						outStr = append(outStr, rv.Index(i).Interface().(string))
+					}
+					vv := reflect.ValueOf(outStr).Convert(t.Field(i).Type)
+					v.Field(i).Set(vv)
+				} else {
+					vv := reflect.ValueOf(data).Convert(t.Field(i).Type)
+					v.Field(i).Set(vv)
+				}
+
```
noboruma, line 280: nit: might be good to have some tests around this function

noboruma (approved):
Happy to merge with the 2 changes proposed

deepfence_utils/utils/utils.go
varunsharma0286: This is a good suggestion.
Will be added in subsequent PR.

PR #1428, opened 2023-07-28
ISSUE-1901: Adding logic to handle the listing of tags correctly (Do not merge yet)

varunsharma0286:
Fixes #
https://github.com/deepfence/enterprise-roadmap/issues/1901

The changes included in this PR handles the following cases:
* When the image has multiple tags in the same repositories.
* When the same image is shared across multiple repositories. 
It can have same or different tags in different repositories.

### DIFF
--- a/deepfence_server/model/registry.go
+++ b/deepfence_server/model/registry.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"mime/multipart"
+	"strings"
 	"time"
 
 	commonConstants "github.com/deepfence/ThreatMapper/deepfence_server/constants/common"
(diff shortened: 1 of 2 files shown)


deepfence_server/model/registry.go
```diff
@@ -389,28 +390,116 @@ func ListImages(ctx context.Context, registryId string, filter reporters.FieldsF
...
+			Add this to the exclusion list "nodeListEx"
+
+		Step3: Iterate the ContainerImage nodes in nodeListEx.
+		Iterate through the tags for this ContainerImage and only keep one's
+		which are in the "stubTags". After adding, delete the tag from the stubTags.
+		Add this ContainerImage to the final list
+	*/
+
+	//Step1: Prepare a map "stubTags" of all the valid tags for this repository.
+	stubTags := make(map[string]bool)
```
noboruma, line 414: Hum.. I fail to see how this will work, `subTags` will end up with tags from a set of images that are not related to each others. We could end up in situation like:
```
Repo A: Image A: Tag A
Repo A: Image B: Tag B
Repo B: Image A: Tag B
```
Now you will end up with `Repo A: Image A: Tag A` which does not exist

noboruma (changes requested):
Logic looks complicated here, what we should have is:
```
m.Props["docker_image_tag_list"] = intersect(m.Props["docker_image_tag_list"], l.Props["tags"])
```
With interesect being a simple:
```
func intersect(a, b []string) []string {
   res := []string{}
   
   for i := range b {
      for j := range a {
         if b[i] == a[j] {
           res = append(res, b[i])
         }
      }
   }
   return res
}
```

The logic being: `m.Props` contains all the tags an `ImageID` can have across *all* repositories. `l.Props` contains all the tags an `ImageName` can have in a *given* repository X, thus intersect of x & l will be all the tags contained associated with an ImageID inside RespostoryX, which is the subset we are after IIRC

varunsharma0286:
> Logic looks complicated here, what we should have is:
> 
> ```
> m.Props["docker_image_tag_list"] = intersect(m.Props["docker_image_tag_list"], l.Props["tags"])
> ```
> 
> With interesect being a simple:
> 
> ```
> func intersect(a, b []string) []string {
>    res := []string{}
>    
>    for i := range b {
>       for j := range a {
>          if b[i] == a[j] {
>            res = append(res, b[i])
>          }
>       }
>    }
>    return res
> }
> ```
> 
> The logic being: `m.Props` contains all the tags an `ImageID` can have across _all_ repositories. `l.Props` contains all the tags an `ImageName` can have in a _given_ repository X, thus intersect of `x & l` will give all the tags associated with an ImageID inside RespostoryX, which is the subset we are after if I am not mistaken

For the above comment:

Lets say we have the below situation:
Repo A -> Image A- > Tag1, Tag2
Repo B -> Image A  -> Tag3
Repo B -> Image B -> Tag1 , Tag4, Tag5

In the DB, the ImageStub for RepoB will have {Tag1, Tag3, Tag4, Tag5} 
The docker_image_tag_list on the Image A will be {Tag1, Tag2, Tag3}

When we take the intersection of Image A for Repo B, we will have the following datasets:
m.Props["docker_image_tag_list"] -> {Tag1, Tag2, Tag3} 
l.Props["tags"] -> {Tag1, Tag3, Tag4, Tag5}.

This will result in the intersection as {Tag1, Tag3} . As we can see, this would be incorrect. 
We need the result only as {Tag3}  (for Repo B+ Image A combination).

deepfence_server/model/registry.go
varunsharma0286: The `stubTags` are set of tags from a single Repository. So for the above example, for Repo A, we will have `stubTags` as {Tag A, Tag B}. And for Repo B, `stubTags` would be {Tag B}.

deepfence_server/pkg/registrysync/sync.go
```diff
@@ -82,13 +82,12 @@ func insertToNeo4j(ctx context.Context, images []model.IngestedContainerImage, r
...
-        n.docker_image_name_list = REDUCE(distinctElements = [], element IN COALESCE(n.docker_image_name_list, []) + row.docker_image_name | CASE WHEN NOT element in distinctElements THEN distinctElements + element ELSE distinctElements END),
+		n.docker_image_tag_list = REDUCE(distinctElements = [], element IN COALESCE(n.docker_image_tag_list, []) + (row.docker_image_name+":"+row.docker_image_tag) | CASE WHEN NOT element in distinctElements THEN distinctElements + element ELSE distinctElements END),
 		n.node_name=n.docker_image_name+":"+n.docker_image_tag,
 		s.updated_at = TIMESTAMP(),
-		s.tags = REDUCE(distinctElements = [], element IN COALESCE(s.tags, []) + n.docker_image_tag_list | CASE WHEN NOT element in distinctElements THEN distinctElements + element ELSE distinctElements END)`,
+		s.tags = REDUCE(distinctElements = [], element IN COALESCE(s.tags, []) + row.docker_image_tag | CASE WHEN NOT element in distinctElements THEN distinctElements + element ELSE distinctElements END)`,
```
noboruma, line 88: nit: we could name it `docker_image_name_tag` to be pedantic

noboruma (approved):
LGTM!

PR #1504, opened 2023-09-04
Spawning multiple goroutine for Notification and Sharing http client

saurabh2253:
This fixes the case when there are all integrations configured and few were not getting processed.

Fixes #
https://github.com/deepfence/enterprise-roadmap/issues/1943

- Add timeout to http requests
- Spawn a goroutine for each integration
- Share http client across all goroutines.


deepfence_worker/cronjobs/notification.go
```diff
@@ -30,24 +30,34 @@ func SendNotifications(msg *message.Message) error {
...
-			processIntegration[model.Secret](msg, integrationRow)
-		case utils.ScanTypeDetectedNode[utils.NEO4J_MALWARE_SCAN]:
-			processIntegration[model.Malware](msg, integrationRow)
-		case utils.ScanTypeDetectedNode[utils.NEO4J_COMPLIANCE_SCAN]:
-			processIntegration[model.Compliance](msg, integrationRow)
-			// cloud compliance scans
-			integrationRow.Resource = utils.ScanTypeDetectedNode[utils.NEO4J_CLOUD_COMPLIANCE_SCAN]
-			processIntegration[model.CloudCompliance](msg, integrationRow)
-		}
+		go processIntegrationRow(wg, integrationRow, msg)
```
noboruma, line 36: I am not sure we need to do this at all.
What is this solving?

noboruma: I would go one step further and limit to 1, removing `go` routine entirely ;)
Unless we have a good reason, here it feels we are making the code more complex for little benefit, we are actually stressing the DB access & memory usage more than necessary.
Instead we should optimize the sending of HTTP requests (batch + go routines)

gnmahanth: ok that should minimize frequent neo4j access

noboruma:
I have found a couple of issues (memory leak, broken batching) in the current code base.
Let's merge this in and revamp the code to make it more generic & fix those issues

PR #1511, opened 2023-09-05
Make sure cron jobs are never returning errors

noboruma:


deepfence_worker/worker.go
```diff
@@ -285,55 +285,58 @@ func startWorker(wml watermill.LoggerAdapter, cfg config) error {
 	worker := NewWorker(wml, cfg, mux)
 
 	// sbom
-	worker.AddNoPublisherHandler(utils.ScanSBOMTask, sbom.NewSBOMScanner(ingestC).ScanSBOM, false)
+	worker.AddNoPublisherHandler(utils.ScanSBOMTask, LogErrorWrapper(sbom.NewSBOMScanner(ingestC).ScanSBOM), false)
+
 	worker.AddHandler(utils.GenerateSBOMTask, sbom.NewSbomGenerator(ingestC).GenerateSbom,
 		utils.ScanSBOMTask, publisher)
```
noboruma, line 291: @gnmahanth is this function safe?
I see it returns errors

gnmahanth: add wrapper here

gnmahanth (approved):

shyam-dev:
We should have a way to propagate relevant errors back to the UI/users

noboruma:
Very good point @shyam-dev 
We should have an error page on the console to make it easy to checks errors & report them if needed

PR #1521, opened 2023-09-06
update server and worker entrypoint to wait for file server

gnmahanth:
Fixes #

- update server and worker entrypoint to wait for file server


ramanan-ravi (approved):

noboruma:
Just to make sure - scheduler will also benefit from that because it's part of worker

gnmahanth:
the Dockerfile and image are same for worker, scheduler and ingester so its handled

Issue #1512, opened 2023-09-05
Enable support for go binary cataloger in package scanner


ramanan-ravi:
Optionally enable support for go binary cataloger in package scanner.
It is currently not enabled because of the high memory consumption.

PR #1524, opened 2023-09-07
added binaries scan for vulnerabilities

milan-deepfence:
Fixes https://github.com/deepfence/ThreatMapper/issues/1512

- added to allow user to inlucde binaries in vulnerability scan
- for registry default select binaries when languages are selected


shyam-dev:
The binary scan option should be presented only when golang is chosen

manV (approved):

PR #1533, opened 2023-09-08
fix watermill middlewares not ack messages after processing

gnmahanth:
Fixes #

- fix watermill middlewares not ack messages after processing error cases
- Enable retry of failed tasks if all retries fail message is acknowledge 
- if task panics the panic is captured and message is acknowledged

### DIFF
--- a/deepfence_worker/utils/watermill.go
+++ b/deepfence_worker/utils/watermill.go
@@ -1,0 +1,137 @@
+package utils
+
+import (
+	"fmt"
+	"runtime/debug"
+
+	"context"
+	"time"
+
+	"github.com/pkg/errors"
+
+	"github.com/cenkalti/backoff/v3"
+
+	"github.com/ThreeDotsLabs/watermill"
+	"github.com/ThreeDotsLabs/watermill/message"
+)
+
+// RecoveredPanicError holds the recovered panic's error along with the stacktrace.
+type RecoveredPanicError struct {
+	V          interface{}
+	Stacktrace string
+}
+
+func (p RecoveredPanicError) Error() string {
+	return fmt.Sprintf("panic occurred: %#v, stacktrace: \n%s", p.V, p.Stacktrace)
+}
+
+// Recoverer recovers from any panic in the handler and appends RecoveredPanicError with the stacktrace
+// to any error returned from the handler.
+func Recoverer(h message.HandlerFunc) message.HandlerFunc {
+	return func(event *message.Message) (events []*message.Message, err error) {
+		panicked := true
+
+		defer func() {
+			if r := recover(); r != nil || panicked {
+				err = errors.WithStack(RecoveredPanicError{V: r, Stacktrace: string(debug.Stack())})
+				// ack message as we don't want to execute panic message again
+				event.Ack()
+			}
+		}()
+
+		events, err = h(event)
+		panicked = false
+		return events, err
+	}
+}
+
+// Retry provides a middleware that retries the handler if errors are returned.
+// The retry behaviour is configurable, with exponential backoff and maximum elapsed time.
+type Retry struct {
+	// MaxRetries is maximum number of times a retry will be attempted.
+	MaxRetries int
+
+	// InitialInterval is the first interval between retries. Subsequent intervals will be scaled by Multiplier.
+	InitialInterval time.Duration
+	// MaxInterval sets the limit for the exponential backoff of retries. The interval will not be increased beyond MaxInterval.
+	MaxInterval time.Duration
+	// Multiplier is the factor by which the waiting interval will be multiplied between retries.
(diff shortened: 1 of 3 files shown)


deepfence_worker/utils/watermill.go
```diff
@@ -0,0 +1,137 @@
...
+
+		expBackoff := backoff.NewExponentialBackOff()
+		expBackoff.InitialInterval = r.InitialInterval
+		expBackoff.MaxInterval = r.MaxInterval
+		expBackoff.Multiplier = r.Multiplier
+		expBackoff.MaxElapsedTime = r.MaxElapsedTime
+		expBackoff.RandomizationFactor = r.RandomizationFactor
+
+		ctx := msg.Context()
+		if r.MaxElapsedTime > 0 {
```
varunsharma0286, line 89: Does this means that the total time for any task is limited by this "MaxElapsedTime" variable. 
From what i see, if the total time taken for processing a task is above 1 minute, we will not retry (even if num retries < 3). 
If that is the case, than i believe 1 minute limit (set in worker.go) will be an issue. 
Reason being, we might have some tasks that will run over 1 minute (like what we had for integrations).

ramanan-ravi (approved):

deepfence_worker/utils/watermill.go
ramanan-ravi: We are setting `MaxElapsedTime: 0`
So there is no time limit, only count based limit.

varunsharma0286: Hmm.. I misread "MaxInterval" with "MaxElapsedTime".

varunsharma0286 (approved):
+1

deepfence_worker/utils/watermill.go
gnmahanth: - currently `MaxElapsedTime` is `0` which is disabled

deepfence_worker/utils/watermill.go
```diff
@@ -0,0 +1,137 @@
...
+
+func (p RecoveredPanicError) Error() string {
+	return fmt.Sprintf("panic occurred: %#v, stacktrace: \n%s", p.V, p.Stacktrace)
+}
+
+// Recoverer recovers from any panic in the handler and appends RecoveredPanicError with the stacktrace
+// to any error returned from the handler.
+func Recoverer(h message.HandlerFunc) message.HandlerFunc {
+	return func(event *message.Message) (events []*message.Message, err error) {
+		panicked := true
```
noboruma, line 32: I don't think we need this?

gnmahanth: It is needed in case if any of the handler cause panic, this prevents other handlers from crashing along with the panic, others keep running and panic is recovered

deepfence_worker/utils/watermill.go
```diff
@@ -0,0 +1,137 @@
...
+		expBackoff.Reset()
+	retryLoop:
+		for {
+			waitTime := expBackoff.NextBackOff()
+			select {
+			case <-ctx.Done():
+				return producedMessages, err
+			case <-time.After(waitTime):
+				// go on
+			}
```
noboruma, line 105: I assume all watermill middleware are running on new goroutines?

deepfence_worker/utils/watermill.go
noboruma: That's weird, panic unwinds the stack, so a `panic()` should be bound to a specific `recover()`

noboruma: Looks like there was a misunderstanding - I was talking about the boolean, not the whole function :)

deepfence_worker/worker.go
```diff
@@ -349,23 +350,23 @@ func startWorker(wml watermill.LoggerAdapter, cfg config) error {
...
+// func LogErrorsWrapper(wrapped func(*message.Message) ([]*message.Message, error)) func(*message.Message) ([]*message.Message, error) {
+// 	return func(msg *message.Message) ([]*message.Message, error) {
+// 		msgs, err := wrapped(msg)
+// 		if err != nil {
+// 			log.Error().Msgf("Cron job err: %v", err)
+// 			return nil, nil
+// 		}
+// 		return msgs, nil
+// 	}
+// }
```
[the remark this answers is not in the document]
gnmahanth: done

deepfence_worker/utils/watermill.go
noboruma: Since we can remove the bool, it means we are indeed executing on diff go-routines. All clear

gnmahanth: yes

noboruma (approved):
Great stuff, that's exactly what we needed

PR #1538, opened 2023-09-09
FIX-INTEGRATIONS: Changes to fix the load issue in integrations

varunsharma0286:
Fixes #

-
-
-

### DIFF
--- a/deepfence_server/pkg/integration/splunk/splunk.go
+++ b/deepfence_server/pkg/integration/splunk/splunk.go
@@ -5,13 +5,21 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/deepfence/ThreatMapper/deepfence_utils/utils"
 	"net/http"
 	"strings"
+	"sync"
 	"time"
+
+	"github.com/deepfence/ThreatMapper/deepfence_utils/utils"
+	"github.com/rs/zerolog/log"
 )
 
 var MaxContentLength = 10000
+
+// HECEvent represents an event for the Splunk HTTP Event Collector (HEC) API
+type HECEvent struct {
+	Event map[string]interface{} `json:"event"`
+}
 
 func New(ctx context.Context, b []byte) (Splunk, error) {
 	var s Splunk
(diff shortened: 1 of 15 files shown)


Gdakgdak22 (approved):

Gdakgdak22 (approved):

deepfence_server/pkg/integration/splunk/splunk.go
```diff
@@ -23,64 +31,93 @@ func New(ctx context.Context, b []byte) (Splunk, error) {
...
+		numRoutines = 1
+	} else if numRoutines > 10 {
+		numRoutines = 10
+	}
+
+	senderChan := make(chan []byte, 500)
+	var wg sync.WaitGroup
+	for i := 0; i < numRoutines; i++ {
+		wg.Add(1)
+		go s.Sender(senderChan, &wg)
```
noboruma, line 53: How much performance gain are we getting with those go routines?

deepfence_server/pkg/integration/splunk/splunk.go
```diff
@@ -23,64 +31,93 @@ func New(ctx context.Context, b []byte) (Splunk, error) {
...
 		fmt.Println("Failed to unmarshal message for splunk", err)
 		return err
 	}
+
+	numRoutines := (len(msg) / 10)
+	if numRoutines == 0 {
+		numRoutines = 1
+	} else if numRoutines > 10 {
+		numRoutines = 10
+	}
```
noboruma, line 47: Could simplify with a one-liner:
```suggestion
   numRoutines := max(min(len(msg)/10, 1),10)
```

deepfence_server/pkg/integration/splunk/splunk.go
```diff
@@ -23,64 +31,93 @@ func New(ctx context.Context, b []byte) (Splunk, error) {
...
+			defer resp.Body.Close()
 
-	// Check the response status code
-	if resp.StatusCode != http.StatusOK {
-		fmt.Println("Failed to send data to Splunk", resp.Status)
+			// Check the response status code
+			if resp.StatusCode != http.StatusOK {
+				log.Info().Msgf("Failed to send data to Splunk %s", resp.Status)
+			}
+		}
```
noboruma, line 121: It will be cleaner (less indentation) to have this code outside the `select`
Since you have only 1 branch, we can simply have:
```
var data any
var ok bool
select {
case data, ok = <-in:
}
...
```

deepfence_worker/cronjobs/cloud_compliance.go
```diff
@@ -320,12 +330,7 @@ func CachePostureProviders(msg *message.Message) error {
 	if err != nil {
 		return err
 	}
-
-	rdb, err := directory.RedisClient(ctx)
-	if err != nil {
-		return err
-	}
-	err = rdb.Set(ctx, constants.RedisKeyPostureProviders, string(postureProvidersJson), 0).Err()
+	err = rdb.Set(ctx, constants.RedisKeyPostureProviders, string(postureProvidersJson), time.Hour*2).Err()
```
noboruma, line 333: Given that this job is executed on a regular basis, we don't need to set an expiry on those entries.
If something needs to be cleared up, let's do it in that job.

deepfence_worker/cronjobs/notification.go
```diff
@@ -31,62 +36,39 @@ func SendNotifications(msg *message.Message) error {
...
-						ErrorMsg: sql.NullString{
-							String: "",
-							Valid:  false,
-						},
-					}
-				}
-				pgClient.UpdateIntegrationStatus(ctx, params)
-			}
-		}(integrationRow)
+		go processIntegrationRow(wg, integrationRow, msg)
```
noboruma, line 46: This code was pushed recently - why is that being reverted?

noboruma (changes requested):
We need to rebase

deepfence_server/pkg/integration/splunk/splunk.go
```diff
@@ -23,64 +31,93 @@ func New(ctx context.Context, b []byte) (Splunk, error) {
...
+				log.Info().Msgf("Failed to create HTTP request: %v", err)
+				continue SenderLoop
+			}
+			req.Header.Set("Authorization", authToken)
+			resp, err := s.client.Do(req)
+			if err != nil {
+				log.Info().Msgf("Failed to send data to Splunk: %v", err)
+				continue SenderLoop
+			}
+			defer resp.Body.Close()
```
noboruma, line 115: We have a leak here, `defer` will only be executed at the end of the function, not at the end of the loop

deepfence_server/pkg/integration/splunk/splunk.go
varunsharma0286: @noboruma We will not be flooding the remote servers as we can have maximum of 40 connections at any given time. 
Reason being we have max of 4 routines for an integration(each for vulnerability, secret, malware, compliance). For each routine, we have max of 10 go routines. That's one of the reason for limiting the routines to 10.

As for the performance, please find below the numbers (time is in milliseconds)
**Base Case**: Logs when the number of routines = 1 
Mon, 11 Sep 2023 04:18:13 +0000 INF splunk/splunk.go:61 SPLUNK::SendNotification complete, numRoutines:1
Mon, 11 Sep 2023 04:18:13 +0000 INF cronjobs/notification.go:221 Notification sent Vulnerability scan 87 messages using splunk id 3, time taken:21644

**With Changes** : Logs when the number of routines = 8
Mon, 11 Sep 2023 04:27:48 +0000 INF splunk/splunk.go:59 SPLUNK::SendNotification complete, numRoutines:8
Mon, 11 Sep 2023 04:27:48 +0000 INF cronjobs/notification.go:221 Notification sent Vulnerability scan 87 messages using splunk id 3, time taken:3418

From the numbers, we can see the speedup of around 6 times.

ramanan-ravi: @varunsharma0286 Can you check you are not getting 429 error in all integrations with this change?
If no 429 error, we are good.

deepfence_worker/cronjobs/cloud_compliance.go
ramanan-ravi: Already fixed in main

deepfence_server/pkg/integration/splunk/splunk.go
varunsharma0286: @ramanan-ravi No 429 :)
Also, these changes are specifically for "Splunk" and "Teams"

deepfence_worker/cronjobs/cloud_compliance.go
noboruma: Just realized it's part of the rebase 👍

deepfence_server/pkg/integration/splunk/splunk.go
noboruma: @varunsharma0286 sounds good, one last verification, are those logs generated with the same code but only change is `numRoutines = 1`?

varunsharma0286: Yups... that's correct.

deepfence_server/pkg/integration/splunk/splunk.go
varunsharma0286: Done

deepfence_worker/cronjobs/notification.go
varunsharma0286: Done.

deepfence_server/pkg/integration/splunk/splunk.go
varunsharma0286: Copy paste :( 
Fixed this.

varunsharma0286:
Rebased the branch.

varunsharma0286:
> We need to rebase

Done.

deepfence_server/pkg/integration/splunk/splunk.go
noboruma: Not for this PR but we would benefit having a pool across all the integrations, simplifying the code and making it easier to maintain

deepfence_server/pkg/integration/teams/teams.go
```diff
@@ -35,35 +38,51 @@ func (t Teams) FormatMessage(message map[string]interface{}, position int) strin
...
 	}
+
+	defer func() {
+		close(senderChan)
+		wg.Wait()
+		log.Info().Msgf("Teams::SendNotification complete, numRoutines:%d", numRoutines)
+	}()
+
+	startIndex := 0
+	endIndex := BatchSize
```
noboruma, line 72: Are we guaranteed that `len(msg)` is always a multiple of `BatchSize`?
If not, this can panic

deepfence_server/pkg/integration/teams/teams.go
```diff
@@ -35,35 +38,51 @@ func (t Teams) FormatMessage(message map[string]interface{}, position int) strin
...
 	return nil
 }
 
-func (t Teams) sendNotification(payloads []map[string]interface{}) error {
+func (t Teams) enqueueNotification(payloads []map[string]interface{},
+	senderChan chan *Payload) {
+
 	message := ""
 	for index, msgMap := range payloads {
 		message += t.FormatMessage(msgMap, index+1)
```
noboruma, line 88: Now I am curious - do we know how long this method takes?
This might be another bottleneck

deepfence_worker/cronjobs/notification.go
```diff
@@ -208,12 +232,21 @@ func processIntegration[T any](msg *message.Message, integrationRow postgresql_d
...
-			integrationRow.Resource, len(results), integrationRow.IntegrationType, integrationRow.ID)
+		log.Info().Msgf("Notification sent %s scan %d messages using %s id %d, time taken:%d",
+			integrationRow.Resource, len(results), integrationRow.IntegrationType,
+			integrationRow.ID, time.Since(profileStart).Milliseconds())
 	}
+	log.Info().Msgf("%s Total Time taken for integration %s: %d", integrationRow.Resource,
+		integrationRow.IntegrationType, time.Since(startTime).Milliseconds())
+	log.Info().Msgf("Time taken for neo4j_query_1: %d", neo4j_query_1)
+	log.Info().Msgf("Time taken for neo4j_query_2: %d", totalQueryTime)
+	log.Info().Msgf("Time taken for sending data : %d", totalSendTime)
```
noboruma, line 250: Let's use `Debug()` instead of `Info()`

deepfence_server/pkg/integration/teams/teams.go
varunsharma0286: I don't think that is the case. In my testing, majority of time (> 99%) was going into the http send.

deepfence_server/pkg/integration/teams/teams.go
varunsharma0286: Hmmm... good catch. 
Fixed this

noboruma (approved):

Issue #1487, opened 2023-08-30
[V2] Send Notification for selected scan results


ramanan-ravi:
Add "Notify" button here, which will send the scan results to all configured integrations. There should be a option to send them as individual items or one (`n` Jira tickets vs 1 ticket)

![image](https://github.com/deepfence/ThreatMapper/assets/38394463/8cb0761c-dc72-4fb7-8c6a-c2a3d6fb9c8d)

PR #1576, opened 2023-09-15
added notify features

milan-deepfence:
Fixes https://github.com/deepfence/ThreatMapper/issues/1487

- Add notify features for vulnerability, secret, malware and posture


manV (commented):
Why did we make notify action with color green? I think we should have it in default color.

manV (approved):

PR #1580, opened 2023-09-15
Add ingester circuit breaker

noboruma:


ramanan-ravi (approved):

deepfence_utils/utils/beatcheck.go
```diff
@@ -0,0 +1,45 @@
+package utils
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"time"
+
+	"github.com/cenkalti/backoff/v4"
+	"go.uber.org/zap"
```
gnmahanth, line 10: change logger to deepfence_utils/log

deepfence_worker/processors/bulk_processor.go
```diff
@@ -3,15 +3,49 @@ package processors
...
+var (
+	wait    chan struct{}
+	breaker sync.RWMutex
+)
+
+func init() {
+	wait = make(chan struct{})
+
+	neo4j_port := os.Getenv("DEEPFENCE_NEO4J_BOLT_PORT")
+	neo4j_host := os.Getenv("DEEPFENCE_NEO4J_HOST")
```
gnmahanth, line 27: can we get `neo4j_host` and `neo4j_port` from `directory` this will break multi tenancy in TS

noboruma: Good point, however the whole process here is based on one tenancy.
We will need to change it for multi-tenancy.

PR #1587, opened 2023-09-19
Clean up & optimize status update queries

noboruma:


gnmahanth (approved):

deepfence_worker/ingesters/common.go
```diff
@@ -36,69 +36,31 @@ func CommitFuncStatus[Status any](ts utils.Neo4jScanType) func(ns string, data [
...
+		WHERE NOT n.status IN $cancel_states
+		SET n.status = row.scan_status,
+			n.status_message = row.scan_message,
+			n.updated_at = TIMESTAMP(),
+			r.` + ingestersUtil.ScanStatusField[ts] + `=n.status,
+			r.` + ingestersUtil.LatestScanIdField[ts] + `=n.node_id,
+			r.` + ingestersUtil.ScanCountField[ts] + `=count(m)`
 
 		recordMap := statusesToMaps(data)
 		in_progress, others := splitInprogressStatus(recordMap)
```
varunsharma0286, line 51: I was thinking about this earlier also. 
We can avoid the `splitInprogressStatus` function. We can do the split in the `statusesToMaps` function. 
From there, it can return 3 slices -> recordMap, in_progress and others.

deepfence_worker/ingesters/common.go
```diff
@@ -36,69 +36,31 @@ func CommitFuncStatus[Status any](ts utils.Neo4jScanType) func(ns string, data [
...
-		if _, err = tx.Run(in_progress_query, map[string]interface{}{
+		log.Debug().Msgf("query: %v", query)
+		if _, err = tx.Run(query, map[string]interface{}{
```
varunsharma0286, line 53: Minor Optimization: We can check for the length of "in_progress" and "others" before running the query.

deepfence_worker/ingesters/common.go
```diff
@@ -36,69 +36,31 @@ func CommitFuncStatus[Status any](ts utils.Neo4jScanType) func(ns string, data [
...
-			WITH n, count(m) as total_count
-			OPTIONAL MATCH (n) -[:DETECTED]- (m)
-			WITH  n, total_count, m.resource as arn, count(m) as count
-			OPTIONAL MATCH (n) -[:SCANNED]- (cn) -[:OWNS]- (cr:CloudResource{arn: arn})
-			SET cn.` + ingestersUtil.ScanCountField[ts] + `=total_count, cn.` + ingestersUtil.ScanStatusField[ts] + `=n.status, cn.` + ingestersUtil.LatestScanIdField[ts] + `=n.node_id
-			SET cr.` + ingestersUtil.ScanCountField[ts] + `=count, cr.` + ingestersUtil.ScanStatusField[ts] + `=n.status, cr.` + ingestersUtil.LatestScanIdField[ts] + `=n.node_id`
-		}
+		query := `
+		UNWIND $batch as row
+		OPTIONAL MATCH (m) -[:DETECTED]- (n:` + string(ts) + `{node_id: row.scan_id}) -[:SCANNED]- (r)
```
varunsharma0286, line 41: We have the node "r" here. Here the "r" can be `ContainerImage` , `Container` etc. 
Now we need to take care of the pod logic if this is `Container` and has `pod_id` field as not null. 
I am not sure if we can implement the logic using the "CASE" statement here and return the pod_id if not null.
Let me know your thoughts on this.

deepfence_worker/ingesters/common.go
noboruma: Both the pod update function & cloud resource update function need the full list, I think it's clearer and more re-usable that way

deepfence_worker/ingesters/common.go
noboruma: This case is still being handled, it is in a separate query (`updatePodScanStatus`)
Later we will do `updateCloudResourceScanStatus` and `updatePodScanStatus` outside the hot path entirely, but for now they are there.

PR #1609, opened 2023-09-25
Fix server controls workload logic

noboruma:


ramanan-ravi (approved):

deepfence_worker/cronjobs/console.go
```diff
@@ -40,6 +40,7 @@ func TriggerConsoleControls(ctx context.Context, t *asynq.Task) error {
 		log.Info().Msgf("Init execute: %v", action.ID)
 		err := ctl.ApplyControl(ctx, action)
 		if err != nil {
+			ScanWorkloadAllocator.Free()
```
gnmahanth, line 43: dont we need to release allocator even in case of error or irrespective of errors

gnmahanth (approved):

deepfence_worker/cronjobs/console.go
noboruma: Only in case of error
When not erroring, the job is responsible for telling when it's done, this can't be decided from the controls.

gnmahanth: ok good

PR #1610, opened 2023-09-25
Fix workload allocator semaphore

noboruma:
Previously we were ignoring the "No enough room error"
This is not needed as long as we correctly update the number of running jobs and decrement the counter accordingly, as per semaphore usage.
This fix addresses that.


gnmahanth (approved):

varunsharma0286 (approved):
+1

Issue #1621, opened 2023-09-26
(feature) Filters in scan results page


ramanan-ravi:
Add more filters in scan results page:
- CVE ID
- Attack vector
- Package
- CVE type (Language) (Was missing, fixed in https://github.com/deepfence/ThreatMapper/pull/1623)

Similarly for other scans

![Vulnerability-Scan-Results-Deepfence](https://github.com/deepfence/ThreatMapper/assets/38394463/4a7200f6-6310-48b6-ba4e-30cfa1eaa8e7)

shyam-dev:
Filter by Language is another good to have item

milan-deepfence:
@shyam-dev @ramanan-ravi @ibreakthecloud  
What filters do we want to add for Malware and Secret?

Result page for Malware scan:
<img width="1651" alt="Screenshot 2023-10-05 at 5 49 09 PM" src="https://github.com/deepfence/ThreatMapper/assets/111341101/b76d29ed-fa38-463d-af41-687b03a04f3f">

Result page for Secret scan:
<img width="1665" alt="Screenshot 2023-10-05 at 5 49 30 PM" src="https://github.com/deepfence/ThreatMapper/assets/111341101/511b5343-3322-41d7-ac59-92cbd2558f79">

ramanan-ravi:
@milan-deepfence 

Nothing in Secret and Malware scans.

In posture scan, we should add *ID*

![image](https://github.com/deepfence/ThreatMapper/assets/38394463/9a25f440-cf0b-41db-9526-7d1d1c67b160)

milan-deepfence:
Need api to get id list with pagination support similar to cveid.

ramanan-ravi:
> Need api to get id list with pagination support similar to cveid.

@ibreakthecloud Is completion api available for posture results?

ibreakthecloud:
ref: https://github.com/deepfence/ThreatMapper/pull/1953

ibreakthecloud:
@milan-deepfence I added the completion api for `compliance` and `cloud-compliance` . You can add filters in posture now.

PR #1640, opened 2023-10-03
add vulnerability completion api

ibreakthecloud:
ref:  #1621

### DIFF
--- a/deepfence_server/reporters/completion/completion.go
+++ b/deepfence_server/reporters/completion/completion.go
@@ -2,7 +2,6 @@ package completion
 
 import (
 	"context"
-	"fmt"
 	"time"
 
 	"github.com/deepfence/ThreatMapper/deepfence_server/model"
@@ -46,14 +45,14 @@ func FieldValueCompletion[T reporters.Cypherable](ctx context.Context, req Compl
 
 	if req.ScanID != "" {
 		query = `
-		MATCH (n{node_id: $scan_id}) -[:DETECTED]-> (m) -[:IS]-> (r:` + dummy.NodeType() + `) ` +
-			getWhereClause("r", req.FieldName, req.Completion) +
-			` RETURN DISTINCT r.` + req.FieldName
+		MATCH (n{node_id: $scan_id}) -[:DETECTED]-> (m) -[:IS]-> (r:` + dummy.NodeType() + `)
+		WHERE n.` + req.FieldName + ` =~ '^` + req.Completion + `.*'
+		RETURN DISTINCT r.` + req.FieldName
 	} else {
 		query = `
-		MATCH (n:` + dummy.NodeType() + `) ` +
-			getWhereClause("n", req.FieldName, req.Completion) +
-			` RETURN DISTINCT n.` + req.FieldName
+		MATCH (n:` + dummy.NodeType() + `) 
+		WHERE n.` + req.FieldName + ` =~ '^` + req.Completion + `.*'
+		RETURN DISTINCT n.` + req.FieldName
 	}
 
 	log.Info().Msgf("completion query: \n%v", query)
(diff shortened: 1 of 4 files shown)


gnmahanth (approved):

deepfence_server/reporters/completion/completion.go
```diff
@@ -43,14 +42,25 @@ func FieldValueCompletion[T reporters.Cypherable](ctx context.Context, req Compl
...
-	query := `
-	MATCH (n:` + dummy.NodeType() + `)
-	WHERE n.` + req.FieldName + ` =~ '^` + req.Completion + `.*'
-	RETURN DISTINCT n.` + req.FieldName
+	query := ""
+
+	if req.ScanID != "" {
+		query = `
+		MATCH (n{node_id: $scan_id}) -[:DETECTED]-> (m) -[:IS]-> (r:` + dummy.NodeType() + `) ` +
+			getWhereClause("r", req.FieldName, req.Completion) +
```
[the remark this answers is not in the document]
ibreakthecloud: so I need to add one more check there in case of empty `completion`, right?

though if we don't put that check it might still work (` WHERE r.cve_type =~ '^.*'`)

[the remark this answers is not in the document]
noboruma: No empty completion is handled already as "everything"
If you look below, empty completion will have the regexp: `^.*`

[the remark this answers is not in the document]
ibreakthecloud: got it!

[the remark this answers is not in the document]
ibreakthecloud: done

gnmahanth (approved):

deepfence_server/reporters/completion/completion.go
```diff
@@ -43,14 +41,29 @@ func FieldValueCompletion[T reporters.Cypherable](ctx context.Context, req Compl
...
+		query = `
+		MATCH (n:` + dummy.NodeType() + `) 
+		WHERE n.` + req.FieldName + ` =~ '^` + req.Completion + `.*'
+		WITH n.` + req.FieldName + ` as ` + req.FieldName + `, n.node_id as node_id ORDER BY node_id
+		RETURN DISTINCT n.` + req.FieldName +
+			req.Window.FetchWindow2CypherQuery()
+	}
 
-	log.Debug().Msgf("completion query: \n%v", query)
+	log.Info().Msgf("completion query: \n%v", query)
```
noboruma, line 62:
```suggestion
	log.Debug().Msgf("completion query: \n%v", query)
```

noboruma (changes requested):
Let's keep debug for queries printing

deepfence_server/reporters/completion/completion.go
```diff
@@ -43,14 +41,29 @@ func FieldValueCompletion[T reporters.Cypherable](ctx context.Context, req Compl
...
-	MATCH (n:` + dummy.NodeType() + `)
-	WHERE n.` + req.FieldName + ` =~ '^` + req.Completion + `.*'
-	RETURN DISTINCT n.` + req.FieldName
+	query := ""
+
+	if req.ScanID != "" {
+		query = `
+		MATCH (n{node_id: $scan_id}) -[:DETECTED]-> (m) -[:IS]-> (r:` + dummy.NodeType() + `)
+		WHERE r.` + req.FieldName + ` =~ '^` + req.Completion + `.*'
+		WITH r.` + req.FieldName + ` as ` + req.FieldName + `, r.node_id as node_id ORDER BY node_id
```
noboruma, line 50: This `WITH` is not needed, you can move it to `RETURN`

ibreakthecloud: I tried with `RETURN`, but since it is with `DISTINCT`  I get error
```
In a WITH/RETURN with DISTINCT or an aggregation, it is not possible to access variables declared before the WITH/RETURN: r (line 3, column 51 (offset: 238))
"                RETURN DISTINCT r.cve_id ORDER BY r.node_id"
```

noboruma: What about doing the `ORDER BY` directly onto the `n.FieldName`?

ibreakthecloud: that should work out! Let me push the change

noboruma (approved):
LGTM

PR #1653, opened 2023-10-10
Posture counts

gnmahanth:
Fixes #

- `scan_count` should be number of accounts which are scanned irrespective of active or inactive accounts 
- trigger `CachePostureProviders` task on detecting compliance scan success, this fixes the delay in sync of cached posture counts
-

### DIFF
--- a/deepfence_worker/cronjobs/cloud_compliance.go
+++ b/deepfence_worker/cronjobs/cloud_compliance.go
@@ -238,7 +238,7 @@ func CachePostureProviders(ctx context.Context, task *asynq.Task) error {
 
 			scan_count_query = `
 			MATCH (n:` + string(neo4jNodeType) + `)
-			WHERE n.pseudo=false
+			WHERE n.pseudo=false and n.agent_running=true
 			MATCH (n) <-[:SCANNED]- (m:` + string(utils.NEO4J_COMPLIANCE_SCAN) + `)
 			RETURN count(distinct n)`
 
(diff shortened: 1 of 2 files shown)


ramanan-ravi (approved):

deepfence_worker/cronjobs/cloud_compliance.go
```diff
@@ -237,9 +238,9 @@ func CachePostureProviders(ctx context.Context, task *asynq.Task) error {
 
 			scan_count_query = `
 			MATCH (n:` + string(neo4jNodeType) + `)
-			WHERE n.pseudo=false and n.active=true and n.agent_running=true
```
jatin-baweja, line 240: Is the condition `n.agent_running=true` not required for linux posture count too?

deepfence_worker/ingesters/common.go
```diff
@@ -156,3 +167,17 @@ func ToMap[T any](data T) map[string]interface{} {
 	_ = json.Unmarshal(out, &bb)
 	return bb
 }
+
+func anyCompleted(data []map[string]interface{}) bool {
+
+	complete := false
+
+	for i := range data {
+		if data[i]["scan_status"].(string) == utils.SCAN_STATUS_SUCCESS {
```
jatin-baweja, line 176: Do we need to check if string conversion gave no error? Otherwise, might run into runtime error

gnmahanth: we dont need to check here as this field is guaranteed to have string data

deepfence_worker/cronjobs/cloud_compliance.go
gnmahanth: hosts with agent_running=false are not allowed to scan so they will not appear here

PR #1660, opened 2023-10-12
Compress s3 data

saurabh2253:
Fixes https://github.com/deepfence/enterprise-roadmap/issues/1965

### DIFF
--- a/deepfence_server/pkg/integration/s3/s3.go
+++ b/deepfence_server/pkg/integration/s3/s3.go
@@ -2,12 +2,13 @@ package s3
 
 import (
 	"bytes"
-	"compress/gzip"
 	"context"
 	"encoding/json"
 	"fmt"
+
 	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
 	"github.com/deepfence/ThreatMapper/deepfence_utils/utils"
+	"github.com/klauspost/compress/gzip"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials"
@@ -21,6 +22,8 @@ func New(ctx context.Context, b []byte) (*S3, error) {
 	if err != nil {
 		return &s, err
 	}
+
+	s.Buffer = new(bytes.Buffer)
 	return &s, nil
 }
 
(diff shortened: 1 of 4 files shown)


deepfence_server/pkg/integration/s3/s3.go
```diff
@@ -2,10 +2,10 @@ package s3
 
 import (
 	"bytes"
+	"compress/gzip"
```
shyam-dev, line 5: consider using https://github.com/klauspost/compress It is a drop in replacement and we have done benchmarks on it previously. It is faster and memory efficient when compared to standard library.

deepfence_server/pkg/integration/s3/s3.go
```diff
@@ -63,14 +63,18 @@ func (s S3) SendNotification(ctx context.Context, message string, extras map[str
 		fmt.Println("Failed to marshal JSON data", err)
 		return err
 	}
-
+	buf := &bytes.Buffer{}
```
shyam-dev, line 66: Can we have a buffer that is always present ? We create it once, use it to compress and then clear it out ?

deepfence_server/pkg/integration/s3/s3.go
```diff
@@ -63,14 +63,18 @@ func (s S3) SendNotification(ctx context.Context, message string, extras map[str
...
+	gzWriter.Write(jsonBytes)
+	gzWriter.Close()
 	// Upload the JSON data to S3
 	svc := s3.New(sess)
 	// Default timeout of aws client is 30 sec
 	_, err = svc.PutObject(&s3.PutObjectInput{
-		Body:   bytes.NewReader(jsonBytes),
-		Bucket: aws.String(s.Config.S3BucketName),
-		Key:    aws.String(s.Config.S3FolderName + "/" + utils.GetDatetimeNow() + ".json"),
+		Body:            bytes.NewReader(buf.Bytes()),
```
shyam-dev, line 74: Can we put in a buffer itself instead of a reader object here ?

deepfence_server/pkg/integration/s3/s3.go
```diff
@@ -63,14 +63,18 @@ func (s S3) SendNotification(ctx context.Context, message string, extras map[str
 		fmt.Println("Failed to marshal JSON data", err)
 		return err
 	}
-
+	buf := &bytes.Buffer{}
+	gzWriter := gzip.NewWriter(buf)
```
varunsharma0286, line 67: Here we can look into the `NewWriterLevel` where we can specify the compression level (which can be configurable).
With the `NewWriter`, we will always use the default compression.

gnmahanth (approved):

PR #1674, opened 2023-10-18
File manager updates

gnmahanth:
Fixes #

-  [fix empty scan id being added to tracking](https://github.com/deepfence/ThreatMapper/commit/be99653101dc5e930fcdbca847c952a37e0c1c13)
- [add bucket name to minio file manager](https://github.com/deepfence/ThreatMapper/commit/e6c861873ecca360f0cec60ba3047f9a9f72df75)

### DIFF
--- a/deepfence_worker/tasks/sbom/generate_sbom.go
+++ b/deepfence_worker/tasks/sbom/generate_sbom.go
@@ -84,6 +84,12 @@ func (s SbomGenerator) GenerateSbom(ctx context.Context, task *asynq.Task) error
 		{Key: "namespace", Value: []byte(tenantID)},
 	}
 
+	log.Info().Msgf("payload: %s ", string(task.Payload()))
+
+	if err := json.Unmarshal(task.Payload(), &params); err != nil {
+		return err
+	}
+
 	res, scanCtx := tasks.StartStatusReporter(params.ScanId,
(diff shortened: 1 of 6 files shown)


ramanan-ravi (approved):

deepfence_worker/tasks/sbom/generate_sbom.go
```diff
@@ -124,6 +116,15 @@ func (s SbomGenerator) GenerateSbom(ctx context.Context, task *asynq.Task) error
...
 
+	log.Info().Msgf("Adding scan id to map:%s", params.ScanId)
+	scanMap.Store(params.ScanId, scanCtx)
+	defer func() {
+		log.Info().Msgf("Removing scan id from map:%s", params.ScanId)
+		scanMap.Delete(params.ScanId)
+		res <- err
+		close(res)
+	}()
+
```
noboruma, line 127: This needs to be right after we acquire `res` line 87 or we will be leaking go routines in case we return before sending to `res`

gnmahanth: in that place the scan id is not known to to as its not yest decoded

gnmahanth: the order was wrong updated in latest commit

PR #1683, opened 2023-10-20
add option to use AWS S3 inplace of existing file server

gnmahanth:
Fixes #

- option to use AWS S3 inplace of existing file server
-
-

### DIFF
--- a/deepfence_utils/utils/utils.go
+++ b/deepfence_utils/utils/utils.go
@@ -455,27 +455,35 @@ func RecursiveZip(pathsToZip []string, excludePathPrefixes []string, destination
 }
 
 func UploadFile(url string, fileName string) ([]byte, int, error) {
-	r, err := os.Open(fileName)
+	// r, err := os.Open(fileName)
+	// if err != nil {
+	// 	return nil, 0, err
+	// }
+	// buf := make([]byte, 512)
+	// _, err = r.Read(buf)
+	// if err != nil {
+	// 	return nil, 0, err
+	// }
+	// r.Close()
+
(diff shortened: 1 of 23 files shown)


ramanan-ravi (approved):

deepfence_utils/utils/utils.go
```diff
@@ -455,27 +455,35 @@ func RecursiveZip(pathsToZip []string, excludePathPrefixes []string, destination
...
+	// r, err := os.Open(fileName)
+	// if err != nil {
+	// 	return nil, 0, err
+	// }
+	// buf := make([]byte, 512)
+	// _, err = r.Read(buf)
+	// if err != nil {
+	// 	return nil, 0, err
+	// }
+	// r.Close()
```
noboruma, line 467: not needed?

deepfence_worker/cronscheduler/init_db.go
```diff
@@ -73,12 +73,12 @@ func initSqlDatabase(ctx context.Context) error {
 	return nil
 }
 
-func InitMinioDatabase() error {
```
noboruma, line 76: Do we want to continue if MinioDatabase init fails?
I think we should exit early if it fails

gnmahanth: will panic on error so that scheduler crashes and we can detect error

deepfence_utils/utils/utils.go
gnmahanth: yes not needed it will removed

ramanan-ravi (approved):

PR #1703, opened 2023-10-31
send root benchmark instead of multiple parent benchmarks

gnmahanth:
Fixes #

- send root benchmark instead of multiple parent benchmarks
-
-


ramanan-ravi (approved):

noboruma (commented):
LGTM

Issue #1711, opened 2023-11-03
async update integration status periodically


ibreakthecloud:

ibreakthecloud:
merged and tested

PR #1717, opened 2023-11-03
backoff for 15 mins for integration failures

ibreakthecloud:
ref: https://github.com/deepfence/ThreatMapper/issues/1711

### DIFF
--- a/deepfence_worker/cronjobs/notification.go
+++ b/deepfence_worker/cronjobs/notification.go
@@ -86,8 +86,10 @@ var fieldsMap = map[string]map[string]string{utils.ScanTypeDetectedNode[utils.NE
 
 const DefaultNotificationErrorBackoff = 15 * time.Minute
 
-var NotificationErrorBackoff time.Duration
-var notificationLock sync.Mutex
+var (
+	NotificationErrorBackoff time.Duration
+	notificationLock         sync.Mutex
+)
 
 func init() {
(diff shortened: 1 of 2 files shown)


deepfence_worker/cronjobs/notification.go
```diff
@@ -103,6 +105,11 @@ func SendNotifications(ctx context.Context, task *asynq.Task) error {
 	wg := sync.WaitGroup{}
 	wg.Add(len(integrations))
 	for _, integrationRow := range integrations {
+		if integrationRow.ErrorMsg.String != "" && time.Since(integrationRow.LastSentTime.Time) < DefaultNotificationErrorBackoff {
```
ramanan-ravi, line 108: This integration will get blocked permanently I think. It needs to recover if the service is back up again.

gnmahanth: also this doesn't handle the case when the new integration is added the field `LastSentTime` will be `0`

varunsharma0286: This should be ok... as the `if` check has an "AND" condition.  
* For a new integration, the `ErrorMsg` will be empty.. so the if check will fail and the notifications will be attempted
* For a blocked integration, the `time.Since(integrationRow.LastSentTime.Time)` will be false after the backoff period. So the blocked notifications will be attempted. In this attempt, if there's no error, the previous `ErrorMsg` will be cleared and the integration will become `Active`

gnmahanth (approved):
LGTM

PR #1722, opened 2023-11-07
fix: convert unix time to utc time string in xlsx reports

gnmahanth:
Fixes #
- convert unix time to utc time string in xlsx reports timestamp columns

### DIFF
--- a/deepfence_worker/tasks/reports/xlsx.go
+++ b/deepfence_worker/tasks/reports/xlsx.go
@@ -145,13 +145,14 @@ func vulnerabilityXLSX(ctx context.Context, params utils.ReportParams) (string, 
 
 	offset := 0
 	for _, nodeScanData := range data.NodeWiseData.ScanData {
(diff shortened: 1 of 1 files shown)


ramanan-ravi (approved):

deepfence_worker/tasks/reports/xlsx.go
```diff
@@ -150,7 +151,7 @@ func vulnerabilityXLSX(ctx context.Context, params utils.ReportParams) (string,
 				log.Error().Err(err).Msg("error generating cell name")
 			}
 			value := []interface{}{
-				nodeScanData.ScanInfo.UpdatedAt,
+				time.UnixMilli(nodeScanData.ScanInfo.UpdatedAt).String(),
```
varunsharma0286, line 154: Minor optimization... 
We can take the time value in a variable as follows:
`updatedAt := time.UnixMilli(nodeScanData.ScanInfo.UpdatedAt).String()`
And use that variable in the calls.

Issue #1592, opened 2023-09-20
[V2] Priority scans


ramanan-ravi:
- Option to run a vulnerability / secret / malware / compliance scan on priority
- If 15 container scans / 1000 registry image scans are scheduled and if the user wants to start one container scan on priority (run this before other scheduled scans), it should be supported

noboruma:
We need to add priority fields across all our controls.
We can use 0-255, 0 being high priority and 255 lowest priority. 
That should give us enough room for future features.

Then when we construct the list of controls/actions for worker/agent, we can sort by that field. We should do that directly inside our neo4j queries where we retrieve actions.
Note that there is already an order based on the type of controls, we need to double check this order still makes sense or if all controls need to be treated equal.

This should require little changes and everything should be doable via neo4j (no asynq involvement). Start scan request might need some new priority fields.

PR #1723, opened 2023-11-07
ISSUE-1592: Adding support for priority scans

varunsharma0286:
Fixes #
https://github.com/deepfence/ThreatMapper/issues/1592

### DIFF
--- a/deepfence_server/ingesters/scan_status.go
+++ b/deepfence_server/ingesters/scan_status.go
@@ -62,6 +62,7 @@ func AddNewScan(tx WriteDBTransaction,
 	scan_id string,
 	node_type controls.ScanResource,
 	node_id string,
+	isPriority bool,
 	action controls.Action) error {
 
 	res, err := tx.Run(fmt.Sprintf(`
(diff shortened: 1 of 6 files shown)


deepfence_server/ingesters/scan_status.go
```diff
@@ -128,14 +129,15 @@ func AddNewScan(tx WriteDBTransaction,
 	}
 
 	if _, err = tx.Run(fmt.Sprintf(`
-		MERGE (n:%s{node_id: $scan_id, status: $status, status_message: "", retries: 0, trigger_action: $action, updated_at: TIMESTAMP(), created_at: TIMESTAMP()})
+		MERGE (n:%s{node_id: $scan_id, status: $status, status_message: "", retries: 0, trigger_action: $action, updated_at: TIMESTAMP(), created_at: TIMESTAMP(), is_priority: $is_priority})
```
noboruma, line 132: nit: This query would gain in visibility having a clear `MERGE {node_id}` and `ON CREATE SET` for the remaining fields

deepfence_server/model/scans.go
```diff
@@ -57,8 +57,9 @@ type ScanFilter struct {
 }
 
 type ScanTriggerCommon struct {
-	NodeIds []NodeIdentifier `json:"node_ids" required:"true"`
-	Filters ScanFilter       `json:"filters" required:"true"`
+	NodeIds    []NodeIdentifier `json:"node_ids" required:"true"`
+	Filters    ScanFilter       `json:"filters" required:"true"`
+	IsPriority bool             `json:"is_priority"`
```
noboruma, line 62: Let's make this `required:"true"` as well

deepfence_worker/cronjobs/scheduled_tasks.go
```diff
@@ -115,7 +115,7 @@ func runSystemScheduledTasks(ctx context.Context, messagePayload map[string]inte
 		return nil
 	}
 
-	scanTrigger := model.ScanTriggerCommon{NodeIds: nodeIds, Filters: model.ScanFilter{}}
+	scanTrigger := model.ScanTriggerCommon{NodeIds: nodeIds, Filters: model.ScanFilter{}, IsPriority: false}
```
noboruma, line 118: Can't we use the request's IsPriority field instead of hard coding `false`?
Better leave the choice to the end user

varunsharma0286: Not sure if this should to be added to the scheduled jobs.

deepfence_server/model/scans.go
varunsharma0286: Not sure about making this as required. 
Making this as mandatory will break the existing api clients. 
Also, we are not sure if this needs to be added to the scheduled jobs.

deepfence_worker/cronjobs/scheduled_tasks.go
ramanan-ravi: This is a cron job, so it should be false.

noboruma (approved):

PR #1729, opened 2023-11-09
Fix issue when docker image id is shorter than 12 characters

ramanan-ravi:


deepfence_server/model/utils.go
```diff
@@ -6,7 +6,13 @@ import (
 
 func DigestToID(digest string) (string, string) {
 	imageID := strings.TrimPrefix(digest, "sha256:")
-	return imageID, imageID[:12]
+	var shortImageID string
+	if len(imageID) > 12 {
```
shyam-dev, line 10: How did we arrive at this number 12 ?

noboruma:
sh256 are guaranteed to be 64 characters long (when base64 encoded, 32 bytes otherwise)
Getting digest id lower than 12 bytes hints at a deeper issue

PR #1748, opened 2023-11-21
ISSUE-2015-1: Fixing the timestamp for notifications, panic in server and compilation

varunsharma0286:
Related #
https://github.com/deepfence/enterprise-roadmap/issues/2015
https://github.com/deepfence/ThreatMapper/pull/1625

### DIFF
--- a/deepfence_server/reporters/scan/notify.go
+++ b/deepfence_server/reporters/scan/notify.go
@@ -88,8 +88,8 @@ func injectNodeData[T any](results []T, common model.ScanResultsCommon,
 			flag := integration.IsMessagingFormat(integrationType)
 			if flag {
 				ts := m["updated_at"].(int64)
-				tm := time.Unix(0, ts*int64(time.Millisecond))
-				m["updated_at"] = tm
+				tm := time.Unix(0, ts*int64(time.Millisecond)).In(time.UTC)
+				m["updated_at"] = tm.Format("02-01-2006 15:04:05 MST")
 			}
 		}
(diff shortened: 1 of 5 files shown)


gnmahanth (approved):

deepfence_server/reporters/scan/notify.go
```diff
@@ -88,8 +88,8 @@ func injectNodeData[T any](results []T, common model.ScanResultsCommon,
 			flag := integration.IsMessagingFormat(integrationType)
 			if flag {
 				ts := m["updated_at"].(int64)
-				tm := time.Unix(0, ts*int64(time.Millisecond))
-				m["updated_at"] = tm
+				tm := time.Unix(0, ts*int64(time.Millisecond)).In(time.UTC)
+				m["updated_at"] = tm.Format("02-01-2006 15:04:05 MST")
```
shyam-dev, line 92: Shouldn't this come from the system timezone currently used by the user ?? @varunsharma0286

varunsharma0286: @shyam-dev We are sending the timestamps in UTC for the below integrations:
Slack, Teams, PagerDuty, Email and Jira. 
On the UI, we are displaying the timeformat in the user's timezone.

shyam-dev:
Any integration that is added, is done using either the browser or the API.
Can we send the current system timezone while adding the integration, and
when we send out the feed, convert it to the appropriate timezone ?

PR #1763, opened 2023-11-23
Fix registry sync status when it was just added

ramanan-ravi:
Fixes: https://github.com/deepfence/enterprise-roadmap/issues/2028


varunsharma0286 (approved):
+1

Issue #1750, opened 2023-11-21
UI: Explicit message would be better for a very first admin user during login


milan-deepfence:
After a new console is created.
Try signing in we will get invalid credentials message because user need to be register first. It would be more clear to have a message like:
You will be able to login only after you registration

PR #1785, opened 2023-11-29
For new installation show message in UI

ibreakthecloud:
ref: #1750

### DIFF
--- a/deepfence_server/model/user.go
+++ b/deepfence_server/model/user.go
@@ -355,12 +355,12 @@ func IsFreshSetup(ctx context.Context) (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	_, err = pgClient.GetUsers(ctx)
+	uc, err := pgClient.CountUsers(ctx)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


deepfence_server/model/user.go
```diff
@@ -350,6 +350,21 @@ func GetUserByEmail(ctx context.Context, email string) (*User, int, *postgresqlD
 	return &user, http.StatusOK, pgClient, nil
 }
 
+func IsFreshSetup(ctx context.Context) (bool, error) {
+	pgClient, err := directory.PostgresClient(ctx)
+	if err != nil {
+		return false, err
+	}
+	_, err = pgClient.GetUsers(ctx)
```
ramanan-ravi, line 358: Recommend to use `pgClient.CountUsers(ctx)`

ramanan-ravi (changes requested):

deepfence_server/model/user.go
ibreakthecloud: pushed

ramanan-ravi (approved):

PR #1792, opened 2023-11-30
Base64 encode the api token

ramanan-ravi:
Sometimes users skip copying the namespace which is part of the api token, because it looks unusual


gnmahanth (approved):
LGTM

PR #1805, opened 2023-12-04
Issue-1624: Change the unix timestamp to Human readable

GirishCodeAlchemy:
Fixes #

- Created utility function to convert the unix timestamp to Human readable format.
- Used Timestampformat function to convert the unix timestamp for created and updated field.

### DIFF
--- a/deepfence_server/reporters/scan/notify.go
+++ b/deepfence_server/reporters/scan/notify.go
@@ -92,9 +92,11 @@ func injectNodeData[T any](results []T, common model.ScanResultsCommon,
 		if _, ok := m["updated_at"]; ok {
 			flag := integration.IsMessagingFormat(integrationType)
 			if flag {
-				ts := m["updated_at"].(int64)
-				tm := time.Unix(0, ts*int64(time.Millisecond)).In(time.UTC)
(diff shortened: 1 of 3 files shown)


deepfence_server/reporters/scan/notify.go
```diff
@@ -92,9 +92,11 @@ func injectNodeData[T any](results []T, common model.ScanResultsCommon,
...
 			flag := integration.IsMessagingFormat(integrationType)
 			if flag {
-				ts := m["updated_at"].(int64)
-				tm := time.Unix(0, ts*int64(time.Millisecond)).In(time.UTC)
-				m["updated_at"] = tm.Format("02-01-2006 15:04:05 MST")
+				if ts, err := utils.TimeStampFormat(m["updated_at"], time.Millisecond); err == nil {
+					m["updated_at"] = ts
+				} else {
+					log.Error().Err(err).Msg("Error formatting timestamp")
+				}
```
noboruma, line 99: This pattern is used 3 times, let's have a common function that can be used like:
```
m["updated_at"] = PrintableTimeStamp(m["updated_at"])
```
If there is an error, we print a timestamp from 1970 (and log a message as proposed)

Issue #1759, opened 2023-11-23
Integration with Amazon Bedrock


ramanan-ravi:
Related: https://github.com/deepfence/ThreatMapper/issues/1733

Add integration with Amazon Bedrock

- Get remediation scripts for compliance misconfigurations
- Get remediation scripts for vulnerabilities, secrets, malware

User can configure LLM (ChatGPT and Amazon Bedrock available now) and then click a button on the vulnerability screen / posture screen to get remediation scripts in various formats - cli, console, Terraform, etc

PR #1813, opened 2023-12-05
Option for user to automatically add Amazon Bedrock integrations

ramanan-ravi:
#1759


deepfence_server/handler/generative_ai_integration.go
```diff
@@ -18,42 +20,106 @@ import (
...
+
+	for _, addModelRequest := range addModelRequests {
+		err = h.AddGenerativeAiIntegrationHelper(ctx, addModelRequest, aes, user, pgClient)
+		if err != nil {
+			log.Warn().Msgf(err.Error())
+			continue
+		}
+	}
+
+	httpext.JSON(w, http.StatusOK, model.MessageResponse{Message: api_messages.SuccessIntegrationCreated})
```
shyam-dev, line 91: Are we checking the return value of this function ?

PR #1827, opened 2023-12-11
Fix container cpu and memory report

noboruma:


ramanan-ravi (approved):

deepfence_agent/tools/apache/scope/probe/docker/container.go
```diff
@@ -397,8 +397,8 @@ func (c *container) GetParent() *report.Parent {
 }
 
 func (c *container) GetNode() *report.TopologyNode {
-	c.RLock()
-	defer c.RUnlock()
+	c.Lock()
```
shyam-dev, line 400: Do we really need to lock and unlock the complete function ? Is there a way where we can lock and unlock *only* when we require it ?

PR #1845, opened 2023-12-19
add warn alarm fields for compliance issues

ibreakthecloud:
add warn_alarm_count and cloud_warn_alarm_count fields for compliance with status `warn` or `alarm` for threatgraph


ramanan-ravi (approved):

noboruma (commented):
Looks good to me, just one question: why are we replacing compliance entirely?

ibreakthecloud:
> Looks good to me, just one question: why are we replacing compliance entirely?

Compliance would still be there, to show the total compliance count in case it's needed later. Adding warn_alarm field for only those which failed i.e. `warn` or `alarm` to show in threatgraph. This is the issue: https://github.com/deepfence/enterprise-roadmap/issues/2012

gnmahanth (approved):

noboruma (approved):

PR #1851, opened 2023-12-22
(fix) scan compare between diff registry account

ibreakthecloud:
There was a problem comparing scan. Even if the scan results were same, the diff would show the result.
Problem was same image from different registry will result in different neo4j `Secret`, `Malware`, etc.

Change the way we match, for example instead of matching scan{basescan} -> secret and comparing with scan{compare_to_scan} -> secret, now we are taking `secretrule` (in case of secret) into account, i.e.

matching scan{basescan} -> secret -> secretrule and comparing with scan{compare_to_scan} -> secret -> secretrule
Since the rule is going to be same.

### DIFF
--- a/deepfence_server/reporters/scan/scan_reporters.go
+++ b/deepfence_server/reporters/scan/scan_reporters.go
@@ -694,18 +694,20 @@ func GetScanResultDiff[T any](ctx context.Context, scanType utils.Neo4jScanType,
 	}
 
 	query = `
-	MATCH (m:` + string(scanType) + `{node_id: $base_scan_id}) -[r:DETECTED]-> (d2)
-	WITH collect(d2) as dset
-	MATCH (n:` + string(scanType) + `{node_id: $compare_to_scan_id}) -[r:DETECTED]-> (d)
-	WHERE NOT d in dset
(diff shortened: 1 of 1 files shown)


ramanan-ravi (approved):

noboruma:
Using the Rule is not the right solution, Rules are too generic and we cannot rely on them for comparison. For instance, a rule that matches all .passwd files, it makes no sense to say having a `/etc/nginx/.passwd` is the same as `/etc/systemd/.passwd`

If we want to support comparison between images across different registries, we need to change the secret `node_id` to account for that so the Secret/... nodes can be shared across scans.

This is something we changed recently actually - @ibreakthecloud is that a recent issue?

ibreakthecloud:
> Using the Rule is not the right solution, Rules are too generic and we cannot rely on them for comparison. For instance, a rule that matches all .passwd files, it makes no sense to say having a `/etc/nginx/.passwd` is the same as `/etc/systemd/.passwd`
> 
> If we want to support comparison between images across different registries, we need to change the secret `node_id` to account for that so the Secret/... nodes can be shared across scans.
> 
> This is something we changed recently actually - @ibreakthecloud is that a recent issue?

Not sure if the issue is occurring now, but sure that's the problem. Since the node_id has registry appended in it, comparison fails for same image if scanned from two different sources.

PR #1861, opened 2024-01-03
sync only dump files from neo4j backup to s3

gnmahanth:
Fixes #

- sync only dump files from neo4j backup to s3, this is required as /backup and /data are pointing to same path on helm chart


ramanan-ravi (approved):

noboruma:
LGTM

PR #1866, opened 2024-01-08
Update dependencies

ramanan-ravi:


deepfence_agent/start_agent.sh
```diff
@@ -86,7 +86,7 @@ launch_deepfenced() {
   until [[ "$(ls /var/log/supervisor/deepfenced* 2>/dev/null | wc -w)" != "0" ]]; do
     sleep 5
   done
-  tail -f /var/log/supervisor/deepfenced*
```
noboruma, line 89: Any reason to redirect stderr to `/dev/null`?
If `tail` fails, it will still be good to know why

ramanan-ravi: I was seeing "file truncated" logs, which were unnecessary.
Let me revert this though.

PR #1884, opened 2024-01-16
fix volume path misconfig on neo4j

gnmahanth:
Fixes #

-  fix volume path misconfig on neo4j
- on kubernetes with current volume config there will be data loss if neo4j is restarted, this change fixes that 
- **backup or take dump of neo4j data before upgrading to this changes and restore data after upgrade**

### DIFF
--- a/deployment-scripts/helm-charts/deepfence-console/templates/database/neo4j/neo4j.yaml
+++ b/deployment-scripts/helm-charts/deepfence-console/templates/database/neo4j/neo4j.yaml
@@ -63,12 +63,16 @@ spec:
           volumeMounts:
             - name: "neo4j"
               mountPath: /data
+              subPath: "data"
             - name: "neo4j"
               mountPath: /logs
+              subPath: "logs"
             - name: "neo4j"
               mountPath: /plugins
+              subPath: "plugins"
             - name: "neo4j"
               mountPath: /backups
+              subPath: "backups"
       {{- with .Values.neo4j.nodeSelector }}
       nodeSelector:
         {{- toYaml . | nindent 8 }}


ramanan-ravi (approved):

noboruma (approved):

deployment-scripts/helm-charts/deepfence-console/templates/database/neo4j/neo4j.yaml
```diff
@@ -63,12 +63,16 @@ spec:
           volumeMounts:
             - name: "neo4j"
               mountPath: /data
+              subPath: "data"
```
noboruma, line 66: This is puzzling, I don't fully get why this is needed

gnmahanth: https://stackoverflow.com/questions/65399714/what-is-the-difference-between-subpath-and-mountpath-in-kubernetes

noboruma: Still don't fully grasp what issues was causing data loss.
The way I understand it is, without subPath, the volume could not be shared across multiple `mountPath` but with subPath it can?

noboruma: After offline discussion, it makes sense.
For record, in docker we used to have 1 volume per mount point but for k8s we have 1 volume shared across all mount points.
Now, without `subPath`, we were attaching the same volume at different hook points, messing up the architecture for each `subPath`

PR #1889, opened 2024-01-19
ISSUE-2095-2: Fixing the logic for getting the exploitable counts

varunsharma0286:
Fixes #
https://github.com/deepfence/enterprise-roadmap/issues/2095


ramanan-ravi (approved):

deepfence_server/reporters/graph/threat_graph.go
```diff
@@ -65,7 +65,7 @@ func (tc *ThreatGraphReporter) GetThreatGraph(ctx context.Context, filter Threat
 					VulnerabilityCount:              nodeInfo[index].VulnerabilityCount,
 					ExploitableVulnerabilitiesCount: nodeInfo[index].ExploitableVulnerabilitiesCount,
 					SecretsCount:                    nodeInfo[index].SecretsCount,
-					ExploitableSecretsCount:         nodeInfo[index].ExploitableSecretsCount,
+					ExploitableSecretCount:          nodeInfo[index].ExploitableSecretCount,
```
noboruma, line 68: Why removing plural here?

varunsharma0286: I have changed the field name in the DB to `exploitable_secret_count`. 
Hence made the name change for this field. 
Let me know if we should go with `exploitable_secrets_count` instead?

varunsharma0286: Changed to use plurals

PR #1895, opened 2024-01-22
add api to list all the host for type(including dead ones)

ibreakthecloud:


deepfence_server/router/router.go
```diff
@@ -258,6 +258,9 @@ func SetupRoutes(r *chi.Mux, serverPort string, serveOpenapiDocs bool, ingestC c
 				r.Post("/host", dfHandler.CompleteHostInfo)
 			})
 
+			r.Route("/filter-hosts", func(r chi.Router) {
```
noboruma, line 261: All filter completion are defined here:
https://github.com/deepfence/ThreatStrykerV2/blob/3ea37f06dbece6d4760f4796444a0d2d50af734d/deepfence_server/router/router.go#L276-L280

Please move yours inside the same namespace.

I see we already provide hosts completion, we should reuse that (add a request field for dead entries) instead of reinventing the same method

ibreakthecloud: alright!

milan-deepfence:
@ibreakthecloud @noboruma We have a flag which tells to search host that are active or inactive. One instance:
{
  "node_filter": {
    "filters": {
      "compare_filter": [],
      "contains_filter": {
        "filter_in": { "pseudo": [false], "active": [true] }
      },
      .......
}

Existing search api serves our purpose, I feel new api is not needed.

ibreakthecloud:
> @ibreakthecloud @noboruma We have a flag which tells to search host that are active or inactive. One instance: { "node_filter": { "filters": { "compare_filter": [], "contains_filter": { "filter_in": { "pseudo": [false], "active": [true] } }, ....... }
> 
> Existing search api serves our purpose, I feel new api is not needed.

Closing this PR since new API is no more needed.

PR #1932, opened 2024-02-05
Bubble up error messages in integration

ibreakthecloud:
- send exact error msgs occured during validation of integration

### DIFF
--- a/deepfence_server/handler/integration.go
+++ b/deepfence_server/handler/integration.go
@@ -77,7 +77,14 @@ func (h *Handler) AddIntegration(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if !obj.IsValidCredential(ctx) {
+	vc, err := obj.IsValidCredential(ctx)
+	if err != nil {
+		log.Error().Msgf("%v", err)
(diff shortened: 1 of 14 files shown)


deepfence_server/handler/integration.go
```diff
@@ -247,7 +254,14 @@ func (h *Handler) UpdateIntegration(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if !obj.IsValidCredential(ctx) {
+	vc, err := obj.IsValidCredential(ctx)
+	if err != nil {
+		log.Error().Msgf("%v", err)
+		h.respondError(&ValidatorError{err: err}, w)
```
ramanan-ravi, line 260: ValidatorError requires a certain format to show error in the specific UI form element.
```
errUserInactive       = ValidatorError{
		err: errors.New("email:user is not active")
	}
```
We can make it `BadDecoding`, which will show error below the form.

ibreakthecloud: makes sense.

ibreakthecloud: done

PR #1933, opened 2024-02-06
refactor logs to add namespace from ctx

gnmahanth:
- refactor logs to add namespace from ctx
-
-

### DIFF
--- a/deepfence_worker/utils/asynq.go
+++ b/deepfence_worker/utils/asynq.go
@@ -30,7 +30,7 @@ func Recoverer(h asynq.Handler) asynq.Handler {
 		defer func() {
 			if r := recover(); r != nil {
 				err = errors.WithStack(RecoveredPanicError{V: r, Stacktrace: string(debug.Stack())})
-				log.Error().Err(err).Msg("recovered from panic")
+				log.Error().Ctx(ctx).Err(err).Msg("recovered from panic")
 			}
 		}()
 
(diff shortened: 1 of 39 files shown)


deepfence_worker/worker.go
```diff
@@ -120,7 +120,7 @@ func NewWorker(ns directory.NamespaceID, cfg wtils.Config) (Worker, context.Canc
 			if retried >= maxRetry {
 				err = fmt.Errorf("retry exhausted for task %s: %w", task.Type(), err)
 			}
-			log.Error().Err(err).Msgf("worker task %s, payload: %s", task.Type(), task.Payload())
+			log.Error().Ctx(ctx).Err(err).Msgf("worker task %s, payload: %s", task.Type(), task.Payload())
```
noboruma, line 123: Can we get rid of that Ctx()?

deepfence_worker/utils/asynq.go
```diff
@@ -30,7 +30,7 @@ func Recoverer(h asynq.Handler) asynq.Handler {
 		defer func() {
 			if r := recover(); r != nil {
 				err = errors.WithStack(RecoveredPanicError{V: r, Stacktrace: string(debug.Stack())})
-				log.Error().Err(err).Msg("recovered from panic")
+				log.Error().Ctx(ctx).Err(err).Msg("recovered from panic")
```
noboruma, line 33: needed?

noboruma (approved):

deepfence_worker/utils/asynq.go
gnmahanth: needed or else it requires re-init log var in those functions

PR #1940, opened 2024-02-07
fix TooManyRequests(429) in sync for dockerhub

ibreakthecloud:


deepfence_server/pkg/df-error/registry.go
```diff
@@ -0,0 +1,7 @@
+package df_error
```
gnmahanth, line 1: Instead of creating new package  can this be moved to deepfence_utils/errors so we can define all common errors in one single place in future

ibreakthecloud: sounds good!

ibreakthecloud: pushed!

gnmahanth (approved):

PR #1952, opened 2024-02-12
fix: missing check if role arn is populated

gnmahanth:
https://github.com/deepfence/enterprise-roadmap/issues/2125


varunsharma0286 (approved):
+1

PR #1971, opened 2024-02-15
Periodically refresh cloud metadata in k8s reporter

ramanan-ravi:


gnmahanth (approved):

deepfence_agent/tools/apache/scope/probe/kubernetes/service.go
```diff
@@ -75,7 +75,7 @@ func (s *service) GetNode() report.TopologyNode {
 	return report.TopologyNode{
 		Metadata: metadata,
 		Parents: &report.Parent{
-			CloudProvider:     cloudProviderNodeId,
+			//CloudProvider:     cloudProviderNodeId,
```
noboruma, line 78: Why are we removing this one?

deepfence_agent/tools/apache/scope/probe/kubernetes/kubernetes_cluster.go
```diff
@@ -10,35 +11,75 @@ import (
...
+
+func (k *kubernetesCluster) GetTopology() report.Topology {
+	k.RLock()
+	defer k.RUnlock()
+
+	return k.k8sTopology
 }
 
 func (k *kubernetesCluster) GetNode() report.TopologyNode {
+	cloudMetadata := cloud_metadata.GetCloudMetadata()
```
noboruma, line 66: We could avoid doing this if we know already we are not "private_cloud"

deepfence_agent/tools/apache/scope/probe/kubernetes/service.go
ramanan-ravi: I don't see it being used in the console.
It is not a global variable anymore.
We can add if required in future.

deepfence_agent/tools/apache/scope/probe/kubernetes/kubernetes_cluster.go
ramanan-ravi: Only after this we know whether we are in a known public cloud or not.

PR #2003, opened 2024-03-05
update neo4j health check to consider database backup

gnmahanth:
Fixes #

-
-
-

### DIFF
--- a/deepfence_neo4j/df.sh
+++ b/deepfence_neo4j/df.sh
@@ -14,6 +14,14 @@ if [ -z "$NEO4J_AUTH" ]; then
     export NEO4J_AUTH=$DEEPFENCE_NEO4J_USER/$DEEPFENCE_NEO4J_PASSWORD
 fi
 
+if [ "$NEO4J_OFFLINE_MAINTENANCE_MODE_ENABLED" = true ]; then
+    while true;
+    do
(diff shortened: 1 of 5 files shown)


ramanan-ravi (approved):

deepfence_neo4j/df.sh
```diff
@@ -14,6 +14,14 @@ if [ -z "$NEO4J_AUTH" ]; then
...
 fi
 
+if [ "$NEO4J_OFFLINE_MAINTENANCE_MODE_ENABLED" = true ]; then
+    while true;
+    do
+        echo 'Neo4j is not running. Remove NEO4J_OFFLINE_MAINTENANCE_MODE_ENABLED env var or set offlineMaintenanceModeEnabled to false to resume normal operation.';
+        sleep 60;
+    done
+fi
+
```
noboruma, line 24: I am not following - this is an infinite loop, right?

gnmahanth: This will be enabled when we want to some maintenance operations on neo4j itself, when set the database will not start also in this mode the kubernetes health check is disabled allowing us to operate on db when process is not running 

similar kind of flag exists in standard neo4j helm chart

noboruma (approved):

PR #2046, opened 2024-03-29
Deploying Deepfence Agents on Linux

ramanan-ravi:
- Add option to deploy Deepfence agent as linux binary
- Add instructions for deploying linux binary agent


gnmahanth (approved):

.gitignore
```diff
@@ -34,6 +34,8 @@ deepfence_frontend/product_version.txt
 
 deepfence_agent/tools/apache/scope/proto/
 deepfence_agent/tools/apache/scope/vendor/
+deepfence_agent/fargate/build
```
noboruma, line 37: nit: I think we should stop using fargate as a name internally, it should be more generic, like host or standalone or binaries

deepfence_agent/binary-install-scripts/start_deepfence_agent.sh
```diff
@@ -0,0 +1,159 @@
+#!/bin/bash
```
noboruma, line 1: Do we need really need a new script?
I would rather keep our one start deepfenced script to avoid misses and future bugs

Issue #1367, opened 2023-07-18
enhancement: test SMTP / test integration buttons


RRcwhiting:
**Additional context**
knowing whether the SMTP integration is working would be nice.

**Describe the solution you'd like**
after setting up SMTP, there should be a "test SMTP" button that just sends a test email.

same probably with slack, teams, etc - just a hardcoded "Test threatmapper notification" boilerplate for each integration.

**Describe alternatives you've considered**
tried generating a scan to see if that would send an email, that's not clear if it did.

**Components/Services**

- [ ] UI/Frontend
- [x] API/Backend
- [ ] Agent
- [ ] Deployment/YAMLs
- [ ] CI/CD Integration
- [ ] Other (specify)

gustysap:
Agree to have the test integration.

And can we have button regarding from email address and smtp username.
Because some SMTP server like sendgrid use **apikey** not email for authentication.
https://docs.sendgrid.com/for-developers/sending-email/integrating-with-the-smtp-api

![Threatmapper - Email Configuration](https://github.com/deepfence/ThreatMapper/assets/30117785/bf8c025f-79ec-40ce-bfda-1255df58a890)

ibreakthecloud:
Hi @gustysap thanks for reporting this. This seems to be like very valid use case. Can you create a new issue for that?

gustysap:
Sure @ibreakthecloud ,

I already created the new issue https://github.com/deepfence/ThreatMapper/issues/2048

Thanks for good fast response

PR #2051, opened 2024-04-01
add api to send test email

ibreakthecloud:
ref: https://github.com/deepfence/ThreatMapper/issues/1367

Adds API to send email to configured email. API needs an email which is receipt email.


ramanan-ravi (approved):

ibreakthecloud:
@milan-deepfence we need UI change for this feature. 
something like this:
image

API needs an email. Add a popup for that

PR #2074, opened 2024-04-09
Dynamically generate signed URL for file server objects

ramanan-ravi:
Fixes https://github.com/deepfence/enterprise-roadmap/issues/2172


noboruma:
We need to cache those pre-signed URL in `redis` so that `deepfence_server` don't spend too much time generating those. My worry is it might be putting extra burden on the file server to generate that many continuously, especially since we request those URLs every 30s per agent, so this won't scale.

With that being said, I am not sure pre-signed URL are a good idea now. Since all our agents are authenticated, there is no need to do all that extra work, we could:
- Rely on JWT instead
- By using JWT, we can adapt the URL and change the domain at the agent level directly, so we don't need to touch URL at all

gnmahanth (approved):

noboruma (approved):

deepfence_utils/threatintel/common.go
```diff
@@ -123,19 +124,26 @@ func DeleteFileMinio(ctx context.Context, fName string) error {
...
-	if err != nil {
-		return "", err
+	var exposedURL string
+	cacheVal := ttlCache.Get(consoleURL + fName)
+	if cacheVal == nil {
+		exposedURL, err = mc.ExposeFile(ctx, fName, false, threatintelPollDuration*3, url.Values{}, consoleURL)
+		if err != nil {
+			return "", err
+		}
+		ttlCache.Set(consoleURL+fName, exposedURL, threatintelPollDuration)
```
noboruma, line 141: This operation is not thread safe - we might have multiple requests accessing the same thing

noboruma (changes requested):
Removing previous approval after checking thread safety

deepfence_utils/threatintel/common.go
noboruma: Actually I confused `set` operation that is not safe with `Set` operation that is safe:
https://github.com/jellydator/ttlcache/blob/v3/cache.go#L313

noboruma (approved):

PR #2085, opened 2024-04-12
Add animation on modal close

milan-deepfence:


milan-deepfence:
Animation works but there is an issue of using it in the App because useFetcher still holds previous state.

PR #2099, opened 2024-04-17
On scan delete, set previous scan id on the nodes

ramanan-ravi:
- Separate functions to delete entire scan and partial scan results
- When a scan is deleted, set previous scan id on the node if it was scanned before. Set empty otherwise.

### DIFF
--- a/deepfence_server/reporters/scan/scan_result_actions.go
+++ b/deepfence_server/reporters/scan/scan_result_actions.go
@@ -298,7 +298,7 @@ func DeleteScan(ctx context.Context, scanType utils.Neo4jScanType, scanID string
 		WITH m
 		OPTIONAL MATCH (s:` + string(scanType) + `) - [:SCANNED] -> (m)
 		WITH max(s.updated_at) as most_recent
-		MATCH (m) <-[:SCANNED]- (s:` + string(scanType) + `{updated_at: most_recent})-[:DETECTED]->(c:Vulnerability)
+		MATCH (m) <-[:SCANNED]- (s:` + string(scanType) + `{updated_at: most_recent})-[:DETECTED]->(c:` + utils.ScanTypeDetectedNode[scanType] + `)
 		WITH s, m, count(distinct c) as scan_count
 		SET m.` + latestScanIDField + `=s.node_id, m.` + scanCountField + `=scan_count, m.` + scanStatusField + `=s.status`
 		log.Debug().Msgf("Query to reset scan status: %v", query)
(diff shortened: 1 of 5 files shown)


gnmahanth (approved):

deepfence_server/reporters/scan/scan_result_actions.go
```diff
@@ -186,32 +186,25 @@ func DeleteScan(ctx context.Context, scanType utils.Neo4jScanType, scanID string
...
-		}
-	} else {
-		_, err = tx.Run(ctx, `
-		MATCH (m:`+string(scanType)+`{node_id: $scan_id})
-		OPTIONAL MATCH (m)-[r:DETECTED]-> (n:`+utils.ScanTypeDetectedNode[scanType]+`)
-		DETACH DELETE m,r`, map[string]interface{}{"scan_id": scanID})
-		if err != nil {
-			return err
-		}
+		DELETE r`, map[string]interface{}{"node_ids": nodeIDs, "scan_id": scanID})
```
noboruma, line 192: Don't we need a `DETACH` here?

deepfence_server/reporters/scan/scan_result_actions.go
```diff
@@ -224,108 +217,128 @@ func DeleteScan(ctx context.Context, scanType utils.Neo4jScanType, scanID string
...
-				return err
-			}
-			runtimeSbomFile := path.Join("/sbom", "runtime-"+utils.ScanIDReplacer.Replace(scanID)+".json")
-			err = mc.DeleteFile(ctx, runtimeSbomFile, true, minio.RemoveObjectOptions{ForceDelete: true})
-			if err != nil {
-				log.Error().Err(err).Msgf("failed to delete runtime sbom for scan id %s", scanID)
-				return err
-			}
-		}
+func getScanNodeID(ctx context.Context, res neo4j.ResultWithContext) (nodeID string, nodeType string) {
```
noboruma, line 224: Let's return an `error` here instead of logging

deepfence_server/reporters/scan/scan_result_actions.go
```diff
@@ -224,108 +217,128 @@ func DeleteScan(ctx context.Context, scanType utils.Neo4jScanType, scanID string
...
+	rec, err := res.Single(ctx)
+	if err != nil {
+		log.Warn().Msg(err.Error())
+		return
 	}
+	if rec.Values[0] != nil && rec.Values[1] != nil {
+		return fmt.Sprintf("%v", rec.Values[0]), fmt.Sprintf("%v", rec.Values[1])
```
noboruma, line 231:
```suggestion
		return rec.Values[0].(string), rec.Values[1].(string)
```
No need to allocate new strings here

noboruma (approved):
LGTM (after addressing comments)

deepfence_server/reporters/scan/scan_result_actions.go
ramanan-ravi: We are deleting the edge here, not the node

PR #2108, opened 2024-04-22
report error when results exceed 100k nodes

gnmahanth:
- report error when results exceed 100k nodes while generating reports 
- add status_message filed in reports data  
- change reports struck in inprogress to error state if in inprogress for greater than 30mins
- [remove html generation when generating pdf reports](https://github.com/deepfence/ThreatMapper/pull/2108/commits/bfd5332ea4dbbc9db9031f7582b8dbfeccfc0ada)


deepfence_worker/tasks/reports/data.go
```diff
@@ -194,11 +202,19 @@ func getVulnerabilityData(ctx context.Context, params sdkUtils.ReportParams) (*I
 		NodeWiseData:   nodeWiseData,
 	}
 
+	log.Info().Msgf("total vulnerability records in NodeWiseData is %d", data.NodeWiseData.RecordCount)
+
+	if data.NodeWiseData.RecordCount > sdkUtils.ReportRecordsMax {
+		return &data, ErrorMaxRecords
+	}
+
```
noboruma, line 210: We should bail off earlier, here we will store more than `sdkUtils.ReportRecordsMax` in memory.
We need to break out of the previous `for` loop.
Shall we also put a limit on the number of scans?

deepfence_worker/tasks/reports/data.go
```diff
@@ -240,11 +258,20 @@ func getMostExploitableVulnData(ctx context.Context, params sdkUtils.ReportParam
 		NodeWiseData:   nodeWiseData,
 	}
 
+	log.Info().Msgf("total most exploitable vulnerability records in NodeWiseData is %d",
+		data.NodeWiseData.RecordCount)
+
+	if data.NodeWiseData.RecordCount > sdkUtils.ReportRecordsMax {
+		return &data, ErrorMaxRecords
+	}
+
```
noboruma, line 267: Do we really want to go through all the `entries` if we reached the max?

noboruma (approved):

ramanan-ravi (approved):

PR #2166, opened 2024-05-28
remove order filters when getting sev counts for scans

gnmahanth:
Fixes #

- remove order filters when getting sev counts for scans
- this also fixes the issue where scan result sev counts become empty while sorting the table

### DIFF
--- a/deepfence_server/handler/scan_reports.go
+++ b/deepfence_server/handler/scan_reports.go
@@ -1454,6 +1454,9 @@ func listScanResultsHandlerWithSeverityCounts[T any](w http.ResponseWriter, r *h
 		return nil, model.ScanResultsCommon{}, severityCounts, err
 	}
 
+	// order filter doesnot have meaning when getting sev counts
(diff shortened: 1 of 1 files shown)


ramanan-ravi (approved):

deepfence_server/handler/scan_reports.go
```diff
@@ -1454,6 +1454,9 @@ func listScanResultsHandlerWithSeverityCounts[T any](w http.ResponseWriter, r *h
 		return nil, model.ScanResultsCommon{}, severityCounts, err
 	}
 
+	// order filter doesnot have meaning when getting sev counts
+	req.FieldsFilter.OrderFilter.OrderFields = []reporters.OrderSpec{}
```
noboruma, line 1458: Why is order filter set in the first place?

ramanan-ravi: It's one api endpoint `/scan/results/vulnerability` which returns `results` and `severity counts`. Order filter required for results but not for `severity counts`

PR #2184, opened 2024-06-05
(feat.) add notification apis phase I

ibreakthecloud:
- adds API for notification system
- api for listing scan with status(running, failed, etc.)
- api for listing registry sync in progress
- api for listing failed integrations 

**Requires UI changes**

### DIFF
--- a/deepfence_server/handler/notification.go
+++ b/deepfence_server/handler/notification.go
@@ -32,7 +32,10 @@ func (h *Handler) GetScansHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// respond with scans
-	httpext.JSON(w, http.StatusOK, scans)
+	err = httpext.JSON(w, http.StatusOK, scans)
+	if err != nil {
+		log.Error().Msgf("Error responding: %v", err)
+	}
 }
 
 func (h *Handler) MarkScansReadHandler(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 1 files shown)


deepfence_server/reporters/notification/scan.go
```diff
@@ -0,0 +1,155 @@
...
+	"github.com/deepfence/ThreatMapper/deepfence_utils/log"
+	"github.com/neo4j/neo4j-go-driver/v5/neo4j"
+)
+
+func GetScans(ctx context.Context, scanTypes []string, statues []string) (model.NotificationGetScanResponse, error) {
+	response := model.NotificationGetScanResponse{}
+	var err error
+	for _, scanType := range scanTypes {
+		switch scanType {
+		case "vulnerability":
```
ramanan-ravi, line 19: I think this can be simplified by having a map like this https://github.com/deepfence/ThreatMapper/blob/main/deepfence_utils/utils/constants.go#L137

gnmahanth (approved):

deepfence_server/reporters/notification/scan.go
ramanan-ravi: @ibreakthecloud Can you also check the lint errors?

ibreakthecloud: only possible changes where using constants, we cannot simplify this since each scan result has to go in different fields of struct. 
I did the changes, please check again

PR #2211, opened 2024-06-20
Update integration with doc links in each integration form

milan-deepfence:
- Add docs links for all of the integration except for AWS Security Hub and Google Chronicle
- Refactor integration code to have better management


deepfence_frontend/apps/dashboard/src/features/integrations/components/integration-form/utils.ts
```diff
@@ -0,0 +1,165 @@
...
+  splunk: 'splunk',
+  sumoLogic: 'sumologic',
+  elasticsearch: 'elasticsearch',
+  googleChronicle: 'googlechronicle',
+  awsSecurityHub: 'aws_security_hub',
+  jira: 'jira',
+  s3: 's3',
+} as const;
+
+export function getIntegratinPrettyName(type: string) {
```
manV, line 22: instead of string type, we can use `(typeof IntegrationType)[keyof typeof IntegrationType]` and use exhaustive check.

deepfence_frontend/apps/dashboard/src/features/integrations/components/useIntegrationTableColumn.tsx
```diff
@@ -658,5 +711,10 @@ export const useIntegrationTableColumn = (
     ];
     return columns;
   }, []);
+
+  // if (!integrationType) {
```
manV, line 715: don't we need this?

deepfence_frontend/apps/dashboard/src/features/integrations/components/IntegrationForm.tsx
```diff
@@ -601,10 +74,23 @@ export const IntegrationForm = ({
     return formData?.config?.use_iam_role === 'true';
   });
 
+  const docLink = IntegrationDocsLinkMap[integrationType];
+
   return (
     <>
       {!data?.success ? (
         <fetcher.Form method="post" className="m-4 overflow-y-auto">
+          {docLink ? (
```
manV, line 83: don't we need to check the length also here, since in original map there are empty strings if there is no url

manV (commented):
Some suggestions, but otherwise looks good.

PR #2229, opened 2024-06-28
support sending scan summary & link

gnmahanth:
related to https://github.com/deepfence/enterprise-roadmap/issues/1988

### DIFF
--- a/deepfence_server/constants/api-messages/success.go
+++ b/deepfence_server/constants/api-messages/success.go
@@ -14,4 +14,5 @@ const (
 	ErrIntegrationTypeCannotBeUpdated = "integration type cannot be updated"
 	ErrIntegrationTypeEmpty           = "integration type cannot be empty"
 	ErrNotificationTypeEmpty          = "notification type cannot be empty"
+	ErrSendSummaryNotSupported        = "notification type doesnot support sending summary links"
 )
--- a/deepfence_server/constants/api-messages/success.go
+++ b/deepfence_server/constants/api-messages/success.go
@@ -14,5 +14,5 @@ const (
 	ErrIntegrationTypeCannotBeUpdated = "integration type cannot be updated"
 	ErrIntegrationTypeEmpty           = "integration type cannot be empty"
 	ErrNotificationTypeEmpty          = "notification type cannot be empty"
-	ErrSendSummaryNotSupported        = "notification type doesnot support sending summary links"
+	ErrSendSummaryNotSupported        = "notification type does not support sending summary links"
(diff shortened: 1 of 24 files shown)


deepfence_server/constants/api-messages/success.go
```diff
@@ -14,4 +14,5 @@ const (
 	ErrIntegrationTypeCannotBeUpdated = "integration type cannot be updated"
 	ErrIntegrationTypeEmpty           = "integration type cannot be empty"
 	ErrNotificationTypeEmpty          = "notification type cannot be empty"
+	ErrSendSummaryNotSupported        = "notification type doesnot support sending summary links"
```
ramanan-ravi, line 17: Typo: does not

ramanan-ravi (approved):

deepfence_server/handler/integration.go
```diff
@@ -44,6 +44,18 @@ func (h *Handler) AddIntegration(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if req.SendSummary {
+		if integration.SupportsSummaryLink(req.IntegrationType) {
+			req.Config["send_summary"] = True
```
ibreakthecloud, line 49: question: why are we getting send_summary in req and then setting "send_summary" in config, can't we directly take this in config from request?

ibreakthecloud: or just keep it in main struct, not inside config would be easier to marshal/unmarshal

deepfence_server/pkg/integration/integration.go
```diff
@@ -58,17 +58,32 @@ func GetIntegration(ctx context.Context, integrationType string, b []byte) (Inte
...
 }
 
+func SupportsSummaryLink(integrationType string) bool {
+	switch integrationType {
+	case constants.Slack,
+		constants.Teams,
+		constants.Email:
+		return true
+	}
+	return false
```
ibreakthecloud, line 78: nit: I think we should maintain a array of supported in constants. feel free to ignore this comment

deepfence_server/handler/integration.go
gnmahanth: this config doesn't apply for all integrations and keeping in main struct would required add a new column in the postgresql table

deepfence_server/pkg/integration/integration.go
gnmahanth: will keep this for now

ibreakthecloud (approved):

PR #2253, opened 2024-07-15
[UI] Posture results grouping

manV:
- New grouped view for cloud posture results


deepfence_frontend/apps/dashboard/src/features/postures/pages/PostureCloudDetailModal.tsx
```diff
@@ -274,10 +294,77 @@ const DetailsComponent = ({
...
+const Content = () => {
+  const [isRemediationOpen, setIsRemediationOpen] = useState(false);
+
+  const {
+    data: { data: cloudPostures },
+  } = useGetComplianceDetails();
+  const data = cloudPostures.length ? cloudPostures[0] : undefined;
+
+  let prefix = '';
+  if (data?.cloud_provider?.toLowerCase().includes('aws')) {
```
milan-deepfence, line 306: We may require such condition check in future, I think it would be better to have own utility. Because we had removed possible string and used enum.

deepfence_frontend/apps/dashboard/src/queries/posture.tsx
```diff
@@ -405,14 +414,138 @@ export const postureQueries = createQueryKeys('posture', {
...
+        const totalStatus = Object.values(result.value.status_counts ?? {}).reduce(
+          (acc, value) => {
+            acc = acc + value;
+            return acc;
+          },
+          0,
+        );
+
+        const cloudComplianceStatus = {
+          alarm: result.value.status_counts?.['alarm'] ?? 0,
```
milan-deepfence, line 527: 'alarm' - Same here because we have removed all string references

deepfence_frontend/apps/dashboard/src/features/postures/pages/PostureCloudDetailModal.tsx
manV: let's decide later if the code is repeated at many places, we will refactor it.

PR #2263, opened 2024-07-22
Fix table pagination width

milan-deepfence:
Pagination width does not cover full width


milan-deepfence:
Closing this because it is known issue. Pagination is not extended to full width is to keep the page visible to user when table content is out of screen.

PR #2331, opened 2024-10-05
Enforce DEEPFENCE_FILE_SERVER_PORT to be a string

psych0d0g:
ConfigMap in version "v1" cannot be handled as a ConfigMap: json: cannot unmarshal number into Go struct field ConfigMap.data of type string

due to the template not quoting the port for an external S3 instance, the configmap will not be able to be applied

add quoting to the value for the port

Fixes #

add ` | quote` to the template that pulls the port number from values


ramanan-ravi:
Thanks for the contribution @psych0d0g

